PR #2, opened 2015-07-21
Added MediaTek MT8173 EVB

m943040028:
Signed-off-by: SY Chiu <sy.chiu@linaro.org>
Tested-by: SY Chiu <sy.chiu@linaro.org> (MT8173-EVB)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ $ repo sync
 * QEMU: default.xml
 * FVP: fvp.xml
 * Hikey: hikey.xml
+* MediaTek MT8173 EVB Board: mt8173-evb.xml
 
 ### 2.2 Branches
 Currently we are only using one branch, i.e, the master branch.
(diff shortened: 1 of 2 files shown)


mt8173_evb.xml
```diff
@@ -0,0 +1,34 @@
...
+
+	<!-- busybox -->
+	<project remote="busybox" path="busybox" name="busybox.git" />
+
+	<!-- MediaTek Download Tool -->
+	<project remote="mediatek" path="mtk_tools" name="mtk_tools.git" />
+
+	<!-- Linux kernel -->
+	<project remote="mtk-kernel" path="linux" name="linux-8173.git" revision="4.0rc1"/>
+	<project remote="mtk-kernel" path="patches-upstream" name="patches-upstream.git" revision="4.0rc1"/>
```
pascal-brand-st-dev: Is there any way to have these 2 gits part of linaro-swg? liniux should be already there, and then we could fork patches-upstream?

README.md
```diff
@@ -62,6 +63,15 @@ Location of files/images mentioned in the link above:
 * ```$HOME/devel/optee/arm-trusted-firmware/build/hikey/release/fip.bin```
 * ```$HOME/devel/optee/out/boot-fat.uefi.img```
 
+## 6. MT8173-EVB
+After getting the source and toolchain, please run:
+
+```
+$ make all flash-image
```
pascal-brand-st-dev: Does it make sense to have `make run` instead of `make flash_image`. It would make the same way to run optee on different platforms

mt8173_evb.xml
m943040028: No problem

README.md
m943040028: I would prefer to keep target `flash-image`, since the action it does is flash an op-tee image to board. To keep syntax sync across different platform, I can add a `run` that depends on `flash-image`. Does this OK for you?

pascal-brand-st-dev: All right

PR #4, opened 2015-08-25
Hikey: use stable strace v4.10 version

pascal-brand-st-dev:
Current version of strace is "master". This means it is
not a stable versionn, and may lead to compilation error
due to incompatible version of the components.

Latest stable version v4.10 is used instead

Change-Id: I0906606f4bfd67c4c4362bf428f30378066dad01

### DIFF
--- a/hikey.xml
+++ b/hikey.xml
@@ -12,7 +12,7 @@
 	<default remote="optee" revision="master" />
 
         <!-- strace -->
-        <project remote="sfnet" path="strace" name="code" />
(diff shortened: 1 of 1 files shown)


pascal-brand-st-dev:
@vchong what do you think of this pull-request?

vchong:
@pascal-brand-st-dev hi and sorry for the delay. Didn't think I've review rights. Anyway, as we'd discussed via email, yes, since we don't know if/when master will break, tagging a stable version is a good idea, so +1 here. Thanks!

We should probably tag the other repos as well, but that's another PR for another day.

PR #19, opened 2016-03-11
Add travis.xml, similar to QEMU (default.xml) but cloning with depth 1

jforissier:
With this manifest, the "repo sync" phase on Travis takes less than one
minute vs. 4-6 minutes before.

travis.xml is created from default.xml and modified as follows:
- Add the clone-depth="1" attribute to all projects (shallow cloning)
- Change the BusyBox reference to the 1_24_0 tag. Previously, a SHA1
was used which would not work with shallow cloning.

Note that I could not get a similar result by simply using
"repo init --depth=1" (cloning would take longer).

### DIFF
--- a/travis.xml
+++ b/travis.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<manifest>
+	<remote name="busybox" fetch="git://busybox.net" />
+	<remote name="linaro-swg" fetch="https://github.com/linaro-swg" />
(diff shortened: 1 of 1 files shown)


pascal-brand-st-dev:
Busybox version is given to be 1_24_0 in the commit message, but is 1_24_1 in travis.xml.

When fixed:

    Reviewed-by: Pascal Brand <pascal.brand@linaro.org>

jenswi-linaro:

jforissier:
Commit message fixed and review tags applied.

PR #21, opened 2016-03-22
Update qemu/fvp/hikey on the generic driver

pascal-brand-st-dev:

### DIFF
--- a/default.xml
+++ b/default.xml
@@ -11,14 +11,13 @@
 	<!-- OP-TEE gits -->
 	<project path="optee_os" name="optee_os.git" />
 	<project path="optee_client" name="optee_client.git" />
-	<project path="optee_linuxdriver" name="optee_linuxdriver.git" />
 	<project path="optee_test" name="optee_test.git" />
 
 	<!-- busybox -->
 	<project remote="busybox" path="busybox" name="busybox.git" revision="dbf5a6da6a4295ce26edd1ce34fde567d19afa02" />
 
 	<!-- Linux kernel -->
-	<project remote="linux" path="linux" name="linux.git" revision="refs/tags/v4.1-rc1"/>
+	<project remote="linaro-swg" path="linux" name="linux.git" revision="optee"/>
 
 	<!-- linaro-swg gits -->
 	<project remote="linaro-swg" path="gen_rootfs" name="gen_rootfs.git" />
--- a/fvp.xml
+++ b/fvp.xml
@@ -5,13 +5,13 @@
 	<remote name="linaro-swg" fetch="https://github.com/linaro-swg" />
 	<remote name="linux" fetch="https://git.kernel.org/pub/scm/linux/kernel/git/torvalds" />
 	<remote name="optee" fetch="https://github.com/OP-TEE" />
+	<remote name="landing-teams" fetch="https://git.linaro.org/landing-teams" />
 
 	<default remote="optee" revision="master" />
 
 	<!-- OP-TEE gits -->
 	<project path="optee_os" name="optee_os.git" />
 	<project path="optee_client" name="optee_client.git" />
-	<project path="optee_linuxdriver" name="optee_linuxdriver.git" />
 	<project path="optee_test" name="optee_test.git" />
 
 	<!-- busybox -->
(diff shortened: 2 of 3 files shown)


default.xml
```diff
@@ -11,14 +11,13 @@
...
 	<project path="optee_client" name="optee_client.git" />
-	<project path="optee_linuxdriver" name="optee_linuxdriver.git" />
 	<project path="optee_test" name="optee_test.git" />
 
 	<!-- busybox -->
 	<project remote="busybox" path="busybox" name="busybox.git" revision="dbf5a6da6a4295ce26edd1ce34fde567d19afa02" />
 
 	<!-- Linux kernel -->
-	<project remote="linux" path="linux" name="linux.git" revision="refs/tags/v4.1-rc1"/>
+	<project remote="linaro-swg" path="linux" name="linux.git" revision="gendrv"/>
```
jenswi-linaro: revision should be optee

fvp.xml
```diff
@@ -25,13 +24,13 @@
 	<project remote="arm" path="edk2" name="edk2.git" revision="refs/tags/v2.1-rc0" />
 
 	<!-- Linux kernel -->
-	<project remote="linux" path="linux" name="linux.git" revision="refs/tags/v3.18-rc1"/>
+	<project remote="linaro-swg" path="linux" name="linux.git" revision="gendrv"/>
```
jenswi-linaro: revision should be optee

hikey.xml
```diff
@@ -40,10 +38,7 @@
 	<project remote="savannah" path="grub" name="grub" />
 
 	<!-- Linux kernel -->
-	<project remote="96b-hk" path="linux" name="linux" revision="hikey-mainline-rebase" />
-	<!-- project remote="96boards" path="linux" name="linux" revision="hikey" /-->
-	<!-- project remote="linux" path="linux" name="linux" /-->
-
+	<project remote="linaro-swg" path="linux" name="linux" revision="hikey_gendrv"/>
```
jenswi-linaro: revision should be optee

jforissier: +1, I will create a PR to add the `optee` node to `hi6220-hikey.dts`

pascal-brand-st-dev:
Updated

jenswi-linaro:

pascal-brand-st-dev:
QEMU: A single test is failing: 7633 (GP specific)

jforissier:
FYI it is also 7633 that was failing on HiKey before I increased the shared memory size.

jenswi-linaro:
It looks like we have two choices, either increase shared memory size for all platforms we intend to do GP testing with or change the shared memory allocator in the driver to use best fit instead of first fit. I guess the shared memory allocator in the old driver used best fit.

Changing to best fit is only a few lines of code in the driver so it's an easy fix.

However, I'm not sure it has to stop merging this.

pascal-brand-st-dev:
Thanks @jforissier. With https://github.com/OP-TEE/optee_os/pull/720,

    Tested-by: Pascal Brand <pascal.brand@linaro.org> (QEMU)
    Tested-by: Pascal Brand <pascal.brand@linaro.org> (QEMU - with Pager)

So either we merge PR 720, either we update the linux driver (but then it requires to update the patches proposed to Linux) and revert https://github.com/OP-TEE/optee_os/pull/709. Either way is fine with me.

Any way, this manifest is OK regarding QEMU.

jforissier:
@jenswi-linaro it's best fit already AFAICT: https://github.com/linaro-swg/linux/blob/optee/drivers/tee/tee_shm_pool.c#L74

jenswi-linaro:
@jforissier yes I noticed that too. :-)

vchong:
@pascal-brand-st-dev  Hi, for hikey.mk, can you also please change the remote for arm-trusted-firmware from 96boards to 96boards-hikey?

pascal-brand-st-dev:
@vchong updated

vchong:
@pascal-brand-st-dev Thanks!

fvp.xml
```diff
@@ -25,13 +24,13 @@
 	<project remote="arm" path="edk2" name="edk2.git" revision="refs/tags/v2.1-rc0" />
```
jenswi-linaro: In the gendrv branch we're using:
`<project remote="landing-teams" path="edk2" name="working/arm/edk2.git" revision="16.01" />`

I just tried it here and it seems to work.

fvp.xml
```diff
@@ -11,7 +11,6 @@
 	<!-- OP-TEE gits -->
 	<project path="optee_os" name="optee_os.git" />
 	<project path="optee_client" name="optee_client.git" />
-	<project path="optee_linuxdriver" name="optee_linuxdriver.git" />
 	<project path="optee_test" name="optee_test.git" />
```
jenswi-linaro: Please add:
`<remote name="landing-teams" fetch="https://git.linaro.org/landing-teams" />`

pascal-brand-st-dev:
updated

jenswi-linaro:
With https://github.com/OP-TEE/build/pull/55
`Tested-by: Jens Wiklander <jens.wiklander@linaro.org> (FVP)`

PR #24, opened 2016-04-04
Release 2.0.0: stable versions

pascal-brand-st-dev:

### DIFF
--- a/hikey_stable.xml
+++ b/hikey_stable.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <manifest>
+  <remote fetch="https://github.com/96boards-hikey" name="96b-hk"/>
   <remote fetch="https://github.com/96boards" name="96boards"/>
   <remote fetch="git://busybox.net" name="busybox"/>
   <remote fetch="https://github.com/linaro-swg" name="linaro-swg"/>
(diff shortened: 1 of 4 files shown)


fvp_stable.xml
```diff
@@ -2,22 +2,22 @@
...
-  <project name="optee_client.git" path="optee_client" revision="refs/tags/1.1.0"/>
-  <project name="optee_linuxdriver.git" path="optee_linuxdriver" revision="refs/tags/1.1.0"/>
-  <project name="optee_os.git" path="optee_os" revision="refs/tags/1.1.0"/>
-  <project name="optee_test.git" path="optee_test" revision="refs/tags/1.1.0"/>
+  <project name="gen_rootfs.git" path="gen_rootfs" remote="linaro-swg" revision="44a44ba7b35e25d576bef9b367e16f471c24be65" upstream="master"/>
+  <project name="linux.git" path="linux" remote="linaro-swg" revision="84c7fe3be96cb570561cbd9926d47f76b93a6fed" upstream="optee"/>
+  <project name="optee_client.git" revision="refs/tags/2.0.0"/>
+  <project name="optee_os.git" path="optee_os" revision="refs/tags/2.0.0"/>
+  <project name="optee_test.git" path="optee_test" revision="refs/tags/2.0.0"/>
+  <project name="working/arm/edk2.git" path="edk2" remote="landing-teams" revision="131aec010f565a747ac51e6a95e30a37b5df15cc" upstream="16.01"/>
```
jenswi-linaro: This is the same as the `16.01` tag used in `fvp.xml`, can't we use the tag instead?

hikey_stable.xml
```diff
@@ -10,22 +11,21 @@
...
-  <project name="optee_client" revision="refs/tags/1.1.0"/>
-  <project name="optee_linuxdriver" revision="refs/tags/1.1.0"/>
-  <project name="optee_os" revision="refs/tags/1.1.0"/>
-  <project name="optee_test" revision="refs/tags/1.1.0"/>
-  <project name="patches_hikey" remote="linaro-swg" revision="7652cfa65046a46e6480ce5ad7262d07b2eca1f6" upstream="master"/>
+  <project name="linux" remote="linaro-swg" revision="84c7fe3be96cb570561cbd9926d47f76b93a6fed" upstream="optee"/>
+  <project name="optee_client" revision="refs/tags/2.0.0"/>
+  <project name="optee_os" revision="refs/tags/2.0.0"/>
+  <project name="optee_test" revision="refs/tags/2.0.0"/>
+  <project name="patches_hikey" remote="linaro-swg" revision="ac50f943be0fb0dbd3c02631b773ac0f12972add" upstream="gendrv"/>
```
vchong: Sorry. Revision needs to be update to "278a0084b0bc37f831029f56c678d52e50fb7e42" due to bug and discussion in https://github.com/OP-TEE/optee_os/issues/727.

jbech-linaro:
From the one I've been mainly involved in testing, the MTK board, this looks fine.

hikey_stable.xml
pascal-brand-st-dev: I'll fix. If possible, when pushing on linaro-swg, can we use PR process, so that I see such changes? Thanks,
Pascal.

jenswi-linaro:

fvp_stable.xml
pascal-brand-st-dev: 16.01 is not a tag, this is a branch. So I would prefer to keep it as it is.

Also, the `stable` files are automatically generated. I only change manually `OP-TEE` and `build` commit version to use the tags. I do not change the others. For example, I do not use the tag version of the ARM-TF above. Let me know if I should do another way (meaning use tags when possible - I would prefer not as it means more manual changes, so is error-prone).

hikey_stable.xml
vchong: Understood! Got a bit trigger-happy on the keyboard during the weekend.. :frowning:

pascal-brand-st-dev:
updated

vchong:
`Reviewed-by: Victor Chong <victor.chong@linaro.org> (HiKey)`

fvp_stable.xml
jenswi-linaro: OK, sorry I thought it was a tag.

Yes, I started to suspect that the files where auto generated in some way and that's good. It's quite tedious to review  though, but I guess we'll have to live with that.

PR #37, opened 2016-10-05
Updates for 2.2.0 release

jbech-linaro:
This should work as a base for the 2.2.0 release. Only last minute minute tweaks should be necessary here, before making the actual tag.


jforissier:

jbech-linaro:
I've pushed a few more commits here, the title of the PR doesn't reflect the reality, since there are now also patches for HiKey Debian. Changes since you last looked Jerome is:
- Stepped up and prepared for 2.2.0 tag for optee_* gits.
- Changed from our own 96board HiKey Debian branch to instead use the official branch from the 96Boards team.

jforissier:
Looks good, PR title updated ;)

jforissier:
@jbech-linaro in your first comment (https://github.com/OP-TEE/manifest/pull/37#issue-181118251) you say this should be merged before making the 2.2.0 tag in optee_os/optee_client? Or did I misunderstand? Because now you have added references to those tags. Or does it mean that we should merge this and create the tags simultaneously (more or less)?

jbech-linaro:
Right, with the 2.2.0 tags added, this could probably be merged and it should just work without any further modifications (since the rest hasn't or shouldn't move from now on).

Do you want me to squash the commits to two commits? One for the rename and the rest in a single commit? Or should I keep patches separately for each platform?

jforissier:
@jbech-linaro in general, the more commits, the better ;) everything is fine as is, except for the fixup patch which should be folded into the parent.
`Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>`

Since you have started to update the _stable XMLs in this PR, would you mind doing the same for the platforms that pass the tests as reported in the change log PR? (https://github.com/OP-TEE/optee_os/pull/1087) That is, those that have "extended" or "standard" next to their ID.

jbech-linaro:
I'll do my best :) Expect a few more commits here. For the ones that I can setup (meaning we have repo manifests), I'll try to do it.

jbech-linaro:
I'll do some squash juggling also. Beside just updating to latest, I've updated the rpi3_stable.xml with the new gen_rootfs commit from Jens also. But when everything is squash into the correct patches (one for each platform). I'll grep for all gits one by one, so ensure that they have the same commit.

jforissier:
Thanks @jbech-linaro!
Regarding mtk8173, I'm not sure we should update _stable since we have no feedback from the maintainer. In this case we can't assume our current master is stable and it's probably best to keep _stable unchanged -- which BTW is actually wrong already, because the last known good configuration for mtk8173 is 2.0.0.

jforissier:
`Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>`

jbech-linaro:

jforissier:
Thanks, I'll merge as soon as the change log is pushed to optee_os and the 2.2.0 tags are created.

PR #39, opened 2016-10-11
travis.xml: checkout QEMU commit c5d128f because master is broken

jforissier:
QEMU master needs the following patch before we can use it again:
https://lists.gnu.org/archive/html/qemu-devel/2016-10/msg01979.html

Change-Id: Ic49fdfd83594b68d566990a411743f027dbd7778


jbech-linaro:

PR #40, opened 2016-10-11
travis.xml: remove clone-depth="1"

jforissier:
Clone depth cannot be used when a revision is specified. It results in:
  error: no such remote ref c5d128ffeb5357df1ea3e6de0c13b3d6a09f6064

Change-Id: I12b501ae2dbe40452814745085009285cc474c36


jbech-linaro:
Oh, that old one .. that has given me some headache in the past.

PR #48, opened 2017-04-03
Add support for AM43xx

glneo:
Signed-off-by: Andrew F. Davis <afd@ti.com>

### DIFF
--- a/am43xx.xml
+++ b/am43xx.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<manifest>
+	<remote name="busybox" fetch="git://busybox.net" />
+	<remote name="linaro-swg" fetch="https://github.com/linaro-swg" />
+	<remote name="optee" fetch="https://github.com/OP-TEE" />
+	<remote name="ti-u-boot" fetch="git://git.ti.com/ti-u-boot" />
+	<remote name="ti-linux" fetch="git://git.ti.com/ti-linux-kernel" />
+
+	<!-- OP-TEE gits -->
+	<project remote="optee" path="optee_os" name="optee_os.git" revision="master" />
+	<project remote="optee" path="optee_client" name="optee_client.git" revision="master" />
+	<project remote="optee" path="optee_test" name="optee_test.git" revision="master" />
+
+	<!-- busybox -->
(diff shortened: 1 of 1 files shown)


am43xx.xml
```diff
@@ -0,0 +1,33 @@
...
+
+	<!-- Hello world TA -->
+	<project remote="linaro-swg" path="hello_world" name="hello_world.git" revision="master" />
+
+	<!-- Filesystem -->
+	<project remote="linaro-swg" path="gen_rootfs" name="gen_rootfs.git" revision="master" />
+
+	<!-- Build -->
+	<project remote="optee" path="build" name="build.git" revision="master">
+		<linkfile src="am43xx.mk" dest="build/Makefile" />
```
jbech-linaro: I've no concerns as such, it's just that @igoropaniuk said that everything is the same same in DRA7xx, it's only that platform flag that is different. If that is true, then it might be worth sharing the same makefile in build.git.

Having that said, I believe the commits and branches are different? So, even though we eventually could share the makefile, it's probably not possible to share manifest, right?

But we need to agree whether or not to use the same Makefile. Do you have any concerns @glneo ?

igoropaniuk: @jbech-linaro 
Actually, my concern was related only to the documentation, because of a big amount of copy/paste. I suggested to leave one readme for all TI platforms, but I didn't have anything against the separation scheme Andrew suggested here, where we have one `ti-common.mk` with common TI stuff + separate Makefiles and repo manifests for each specific TI platform. 

I think it's more convenient to have all specific SoC-related stuff in dedicated files, and we can follow our regular build steps described in the main READMEmd, without playing with additional compilation flags for specifying SoC.

jbech-linaro: OK, sounds good to me. Then it sounds to me like we shall go on and merge this as it is and leave eventual consolidation to a later minor commit (after the release instead)? Agree?

jbech-linaro:

am43xx.xml
igoropaniuk: Sounds good. Also, I'm going to add support of TI AM572x (based on my setup steps yesterday) just after this AM43XX PRs are merged, so then we will have a big picture (support of three different SoCs from the one vendor) and decide what to do next: are there any benefits from having different Mafilefiles/manifests or it's better to move everything into one.

igoropaniuk:

glneo:
*Updated

jbech-optee-adm:
Closing the PR since it has been merged manually.

PR #50, opened 2017-04-07
Stable 2.4.0

jbech-linaro:
So, I've tagged all the gits and I've updated all manifests (which this PR is about). I've added a second patch here on top that makes it possible to let Travis test manifest files. The changes in this PR seems to be OK, all builds [passed](https://travis-ci.org/jbech-linaro/manifest/jobs/219745438) after applying and using the 2.4.0 tag.

As you can see all "upstream" sections has been removed, this is something Pascal started removing on some projects and even though it's useful it's just a pain to maintain it. Instead people can use `$ git branch -a --contains <git>` instead if they want to know where the commit comes from.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,62 @@
+# Travis configuration file for downloading and building OP-TEE according to the
+# instructions in build.git
+language: bash
+
+notifications:
+  email:
+    recipients:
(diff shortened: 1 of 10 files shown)


jenswi-linaro:

PR #53, opened 2017-05-31
qemu: add optee_benchmark app rep

igoropaniuk:
`Signed-off-by: Igor Opaniuk <igor.opaniuk@linaro.org>`


jbech-linaro:
As agreed on the chat, let's start out with only QEMU and later add other when needed. That is good as long as there is no build or test regressions.

jforissier:
Merged with commit subject updated (no need for abbreviations).

PR #57, opened 2017-07-04
qemu_v8: clone DTC source

jforissier:
Ubuntu 1.4.04 LTS, 16.04 LTS and even 17.04 provide the libfdt in
version 1.4.0. This version is too old for the latest QEMU that
requires libfdt in version 1.4.2 or higher.

This change updates qemu_v8.xml so that the qemu/dtc/ source code is
cloned locally and therefore the build does not depend on libfdt.

### DIFF
--- a/qemu_v8.xml
+++ b/qemu_v8.xml
@@ -30,6 +30,7 @@
 
 	<!-- QEMU -->
 	<project remote="qemu" path="qemu" name="qemu.git" />
+	<project remote="qemu" path="qemu/dtc" name="dtc.git" />
 
 	<!-- We're using Linaro SWG for ARM-TF until merged officially -->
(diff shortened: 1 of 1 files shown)


jforissier:
I chose not to update `default.xml` (QEMU) and `travis.xml` because they use older versions of QEMU, so the change is not needed. The DTC project should be added when these manifests are upgraded to use a newer version of QEMU.

igoropaniuk:
Although I'm in favor of using built-in support of git submodules in repo (`sync-s="true"` attribute,which  I've mentioned in the previous PR) :)

jforissier:
@igoropaniuk fine, but do you have a solution to avoid cloning ~138 MiB worth of useless submodules when we need only `dtc` which is ~858 KiB?

igoropaniuk:
@jforissier 
> but do you have a solution to avoid cloning ~138 MiB 

Seems that I missed these details :)

Sure, now I totally agree with your solution

PR #58, opened 2017-07-05
QEMU v7/v8: Use a known working revision

jbech-linaro:
For the 2.5.0 release we found some regressions when running latest on
QEMU. After bisecting we managed to find where the issues where
introduced, they are as follows:

  v8: (e75449a346) target/aarch64: optimize indirect branches
  v7: (8a6b28c7b5) target/arm: optimize indirect branches

There is a fix for that upstream, but apparently that fix does not solve
the problems for TrustZone and therefore we need more time to
investigate it. Because of that we decided to temporarily use the
previous merge commit in QEMU that is working fine on both v7 and v8 and
that is:
  (199e19e) Merge remote-tracking branch 'remotes/mjt/tags/trivial-patches-fetch' into staging

### DIFF
--- a/default.xml
+++ b/default.xml
@@ -28,7 +28,8 @@
 	<project remote="linaro-swg" path="optee_benchmark" name="optee_benchmark.git"/>
 
 	<!-- Temporary use the defined commit since upstream tip didn't work -->
-	<project remote="qemu" path="qemu" name="qemu.git" revision="c5d128ffeb5357df1ea3e6de0c13b3d6a09f6064" />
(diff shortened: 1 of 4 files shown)


jforissier:
Since the selected version is quite recent, I suspect that `<project remote="qemu" path="qemu/dtc" name="dtc.git" />` is needed everywhere (see https://github.com/OP-TEE/manifest/pull/57).

jbech-linaro:
Yes, true, I will fix that. Thanks for catching that.

jforissier:
Thanks.
`Reviewed-by: Jerome Forissier <jerome.forissier@linaro.org>`

PR #70, opened 2017-08-09
travis: Enable hikey960

vchong:
Signed-off-by: Victor Chong <victor.chong@linaro.org>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ env:
   - $REPO_PROJ=default
   - $REPO_PROJ=fvp
   - $REPO_PROJ=hikey
+  - $REPO_PROJ=hikey960
   - $REPO_PROJ=juno
   - $REPO_PROJ=mt8173-evb
   - $REPO_PROJ=qemu_v8


jforissier:

vchong:
Tag added. Thanks!

jforissier:
Travis is failing because it expects to find `hikey960_stable.xml`. I suppose that file has to be created.
(But why would the CI script of this project depend on a stable snapshot? Don't we want to test the current state?)

vchong:
So should I create `hikey960_stable.xml`, or wait for @jbech-linaro to comment if we should test current state instead? The latter seems more appropriate.

jforissier:
Yeah, ideally `hikey960_stable.xml` would have to reflect our next (stable) release, so we'd better not create it now. Let's wait before merging this, until we can clarify the purpose of the `_stable` stuff in `.travis.yml`.

jbech-linaro:
Yeah, I too think it is better to wait, hopefully we get somewhere with Linaro CI also soon ...

jforissier:
@jbech-linaro why is Travis testing `_stable`?

jbech-linaro:
> @jbech-linaro why is Travis testing _stable?

Mainly because updates to manifest are most often done when we are releasing, i.e, this was meant to be a way for you to trigger a sanity build test when changing all manifests for a release. Compare that to our "build.git" where we tend to add patches quite often (and therefore doesn't use _stable).

jforissier:
Yeah. To be honest I don't see the point in having _stable manifests. Much better to use tags to identify what is stable (or branches, as I seem to recall that `repo` doesn't know how to check out a tag).
Then Travis could do what it is supposed to do as a CI tool, i.e. build the current state of the Git tree so that PRs can be tested properly.

jbech-linaro:
I think you are right, it was (hate to say it ...) repo that wasn't able to deal with the tags "outside" the manifest that was the issue which made us create stable manifests. If we can change it suite are needs better, then why not? There is a comment at the bottom [here](https://groups.google.com/forum/#!topic/repo-discuss/IBgCUc4QaKg), which suggests using `refs/tags/<tagname>`. I'm pretty sure I tested that too and I couldn't get it to work. But it could be worth double check. If it works, we could get rid of the stable manifests and just tag everything.

igoropaniuk:
@jbech-linaro I think as we have replaced `hello_world` git to `optee_examples` in all manifests, I suppose we should do what you suggested ASAP (as all makefiles in build.git expect that `<repo>/optee_examples` exists), because all further builds will basically fail.

jbech-linaro:
> @jbech-linaro I think as we have replaced hello_world git to optee_examples in all manifests, I suppose we should do what you suggested ASAP

What did I suggest and what do you suggest to change? :) Didn't we conclude that we couldn't use tags?

jforissier:
@jbech-linaro integration should test `foo.xml` not `foo_stable.xml`.

jbech-linaro:
@jforissier , right if we can get it to work (i.e., check out stable tagged manifests using repo). Igor, please give it a try and if it works, then let's switch at next release. Sounds like a plan?

vchong:
@jforissier Ok to merge this now?

jforissier:
@vchong yes

PR #71, opened 2017-09-01
Switch to upstream arm tf

jenswi-linaro:


jenswi-linaro:
Depends on https://github.com/OP-TEE/build/pull/180

jenswi-linaro:
Added patch for Juno also

jforissier:
Nice!

(I'd rather not have the Change-Id tag btw).

jenswi-linaro:
Change-Id tag removed, Acked-by tag added.

PR #73, opened 2017-09-06
travis.xml: use optee_examples instead of hello_world

jforissier:


jbech-linaro:

PR #77, opened 2017-09-16
benchmark: support for HiKey6220 and TI SoCs(am57xx, dra7xx)

igoropaniuk:
Relates to: https://github.com/OP-TEE/build/pull/194
`Signed-off-by: Igor Opaniuk <igor.opaniuk@linaro.org>`

### DIFF
--- a/am57xx.xml
+++ b/am57xx.xml
@@ -19,6 +19,8 @@
 
 	<!-- Linux kernel -->
 	<project remote="ti-linux" path="linux" name="ti-linux-kernel.git" revision="ti-lsk-linux-4.9.y" />
+	<!-- Benchmark app -->
+	<project remote="linaro-swg" path="optee_benchmark" name="optee_benchmark.git" />
 
 	<!-- Sample applications -->
 	<project remote="linaro-swg" path="optee_examples" name="optee_examples.git" revision="master" />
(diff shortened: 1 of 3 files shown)


jbech-linaro:
Even though I've seen it in several places I think it is HiKey 620 and not HiKey 6220 (it comes from HiSilicon Kirin 620 SoC ), @jforissier, do you agree?
Besides that ...

jforissier:
@jbech-linaro correct (as far as I can tell!).

igoropaniuk:
@jbech-linaro @jforissier Pretty interesting that we are using dts with such filename for HiKey620 build `arch/arm64/boot/dts/hisilicon/hi6220-hikey.dts` (Check https://github.com/OP-TEE/build/blob/master/hikey.mk#L157 ). Also I mentioned while ago the same "HiKey6220" title in LAVA device list.

 I guess that both HiKey620 and HiKey6220 stand for the same physical board, no?

jforissier:
Ah, yes, now I understand. The "marketing" name of the SoC is "Kirin 620", but the technical reference (the chip marking) is "Hi6220". So both designations can be considered OK.

igoropaniuk:
Tag applied and rebased, thanks!

PR #81, opened 2017-10-26
Stable versions of manifests on 2.6.0

jforissier:
Change-Id: I3a45cf9335854913170d467b9f3d108ec1411843

### DIFF
--- a/hikey_stable.xml
+++ b/hikey_stable.xml
@@ -1,28 +1,30 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest>
   <remote fetch="https://github.com/96boards-hikey" name="96b-hk"/>
-  <remote fetch="https://github.com/96boards" name="96boards"/>
+  <remote fetch="https://github.com/ARM-software" name="arm"/>
   <remote fetch="git://busybox.net" name="busybox"/>
   <remote fetch="https://github.com/linaro-swg" name="linaro-swg"/>
   <remote fetch="https://github.com/OP-TEE" name="optee"/>
   <remote fetch="git://git.savannah.gnu.org" name="savannah"/>
   <remote fetch="git://git.code.sf.net/p/strace" name="sfnet"/>
   <default remote="optee" revision="master"/>
-  <project name="arm-trusted-firmware" remote="96b-hk" revision="fb1158a365e2bf5bba638cde950678fddf67fe60"/>
-  <project name="build" revision="refs/tags/2.5.0">
+  <project name="OpenPlatformPkg" remote="96b-hk" revision="8c2f9655ec46036ed7e412defe851b99fa205b75"/>
+  <project name="arm-trusted-firmware" remote="arm" revision="3b39efa49d9949c80b6e76b99829f84619b46110"/>
+  <project name="atf-fastboot" remote="96b-hk" revision="5b0d44c057bc0005965da990e8add72670810996"/>
+  <project name="build" revision="refs/tags/2.6.0">
     <linkfile dest="build/Makefile" src="hikey.mk"/>
   </project>
-  <project name="burn-boot" remote="96boards" revision="db522b114fdec4d06d0a1705501fedc277903453"/>
+  <project name="burn-boot" remote="96b-hk" revision="bee2ea1660f3a03df8d391fb75aa08dbc3441856"/>
   <project name="busybox" remote="busybox" revision="dbf5a6da6a4295ce26edd1ce34fde567d19afa02"/>
-  <project name="code" path="strace" remote="sfnet" revision="735c2ad872047e5d66c28eab2dc1aae31aac96ab"/>
-  <project name="edk2" remote="96boards" revision="7da6cdf18127ee52e79f3cf990c8d630fd037784"/>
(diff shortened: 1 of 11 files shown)


hikey_stable.xml
```diff
@@ -1,28 +1,30 @@
...
-  <project name="burn-boot" remote="96boards" revision="db522b114fdec4d06d0a1705501fedc277903453"/>
+  <project name="burn-boot" remote="96b-hk" revision="bee2ea1660f3a03df8d391fb75aa08dbc3441856"/>
   <project name="busybox" remote="busybox" revision="dbf5a6da6a4295ce26edd1ce34fde567d19afa02"/>
-  <project name="code" path="strace" remote="sfnet" revision="735c2ad872047e5d66c28eab2dc1aae31aac96ab"/>
-  <project name="edk2" remote="96boards" revision="7da6cdf18127ee52e79f3cf990c8d630fd037784"/>
-  <project name="gen_rootfs" remote="linaro-swg" revision="cb8c7736a46d1ff459f159b2115395d504c96b45"/>
+  <project name="code" path="strace" remote="sfnet" revision="69c68ef5bf588fe22f1e76cc6464c70227418da7"/>
+  <project name="edk2" remote="96b-hk" revision="19f735580a73a349e46214f22be5688ad834a334"/>
+  <project name="gen_rootfs" remote="linaro-swg" revision="fe617d470e45778c909038bf3e7ca15174a4f893"/>
   <project name="grub" remote="savannah" revision="e54c99aaff5e5f6f5d3b06028506c57e66d8ef77"/>
```
jbech-linaro: @vchong , do you know why we have `grub` sticking to this particular commit? I've run into issue on HiKey when using newer x86 toolchains and by just checking out `savannah/master` things are fine again. (Having that said, I don't think we should do a last minute change here, i.e., let's keep this for now).

jbech-linaro:
Looking good to me:

hikey_stable.xml
jforissier: FYI, this pinned version of `edk2` won't build with GCC 7 either (Ubuntu 17.04 comes with GCC 7 as the default compiler).

jbech-linaro: That's the other issue I have on my HiKey build on Arch Linux, but since I know HiKey has been using their own forks I didn't want to try other commits, so instead what I've done so far in edk2 is:

```patch
From 6e84b399c5f861086480354967d0ded465c7b255 Mon Sep 17 00:00:00 2001
From: Joakim Bech <joakim.bech@linaro.org>
Date: Wed, 25 Oct 2017 12:35:25 +0200
Subject: [PATCH] Jocke

---
 BaseTools/Source/C/VfrCompile/VfrUtilityLib.cpp | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/BaseTools/Source/C/VfrCompile/VfrUtilityLib.cpp b/BaseTools/Source/C/VfrCompile/VfrUtilityLib.cpp
index 3ca57ed741..4fa066dd9f 100644
--- a/BaseTools/Source/C/VfrCompile/VfrUtilityLib.cpp
+++ b/BaseTools/Source/C/VfrCompile/VfrUtilityLib.cpp
@@ -3372,7 +3372,7 @@ CVfrStringDB::GetVarStoreNameFormStringId (
   UINT8       BlockType;
   EFI_HII_STRING_PACKAGE_HDR *PkgHeader;
   
-  if (mStringFileName == '\0' ) {
+  if (mStringFileName == NULL ) {
     return NULL;
   }
 
-- 
2.14.1
```

jforissier: Yeah, the edk2 issue was fixed 9 month ago [upstream](https://github.com/tianocore/edk2/commit/d37fa01fbbe2), but I guess hikey-96boards don't get the fixes. Maybe they should cherry-pick this one.

vchong: > why we have grub sticking to this particular commit?

https://github.com/OP-TEE/build/issues/144. I guess we can revert this change since you're building fine again with master.

> this pinned version of edk2 won't build with GCC 7

Iinm edk2 currently only supports up to GCC5 build profile. E.g. https://github.com/tianocore/edk2/blob/master/BaseTools/Conf/tools_def.template#L372.

jbech-linaro: Alright, thanks @vchong . I don't want to stir around things now with the release, so I think we should just keep the tags, but as soon as possible we should update this on non-stable so there is at least one less issue to think about.


> Iinm edk2 currently only supports up to GCC5 build profile. E.g.

Yeah, that's good for ... the dinosaurs :)

jforissier:
Merged manually into master.

hikey_stable.xml
vchong: @jbech-linaro Np. https://github.com/OP-TEE/manifest/pull/82 submitted.

vchong: > Yeah, the edk2 issue was fixed 9 month ago upstream, but I guess hikey-96boards don't get the fixes. Maybe they should cherry-pick this one.

@jforissier @jbech-linaro https://github.com/96boards-hikey/edk2/pull/37. Coincidence?

jbech-linaro: I guess one reason could be that Fathi has been trying to enable OP-TEE on various builds lately and I guess he is updating the OE builds now with 2.6.0.

PR #84, opened 2017-11-15
Add grub to the FVP manifest

jbech-linaro:
This is a first step in updating EDK2 on FVP. The new way of doing this will include grub. So we can merge this as a standalone patch and later patches would not need to be merged at the "same" time as this patch.

**EDIT** : Actually I need to push more patches to `manifest.git`, i.e., one for updating `ekd2`, so there will still be patches that are dependant on each other and would need to be merged at the same time. So, the comment above doesn't make that much sense due to that.

### DIFF
--- a/fvp.xml
+++ b/fvp.xml
@@ -4,7 +4,7 @@
 	<remote name="busybox" fetch="git://busybox.net" />
 	<remote name="linaro-swg" fetch="https://github.com/linaro-swg" />
 	<remote name="optee" fetch="https://github.com/OP-TEE" />
-	<remote name="landing-teams" fetch="https://git.linaro.org/landing-teams" />
+	<remote name="tianocore" fetch="https://github.com/tianocore" />
 	<remote name="savannah" fetch="https://git.savannah.gnu.org/git" />
 
 	<default remote="optee" revision="master" />
@@ -21,7 +21,8 @@
 	<project remote="arm" path="arm-trusted-firmware" name="arm-trusted-firmware.git" />
 
 	<!-- Tianocore, EDK2 -->
-	<project remote="landing-teams" path="edk2" name="working/arm/edk2.git" revision="16.01" />
+	<project remote="tianocore" path="edk2" name="edk2.git" />
+	<project remote="tianocore" path="edk2-platforms" name="edk2-platforms.git" />
 
 	<!-- Linux kernel -->
 	<project remote="linaro-swg" path="linux" name="linux.git" revision="optee"/>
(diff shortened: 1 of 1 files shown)


jbech-linaro:
Travis fails and I think the reason is that it cannot find commit:
`Linux kernel: 50403184d40d04b3daf140417e031c16c2985eaf`
Why? I suspect it is because we have overwritten the `optee` branch. I.e., that particular commit is on `optee-v4.9-20171027` but you cannot find it on the branch `optee`. So why does this happen? I guess we got the wrong kernel commit in (most) our xml-files on the v2.6.0 release? I believe that means that you cannot checkout the last stable ...

jforissier:
@jbech-linaro regarding the commit that cannot be found, I think your analysis is correct, repo does not seem to be fetching the tags, only the specified branch (`optee`), hence it does not get the desired SHA1. Yet another limitation of `repo`... :open_mouth: (FYI, there is no such issue with git submodules, I checked it).
The solution may be to replace the SHA1 with `refs/tags/optee-v4.9-20171027` in `*_stable.xml`.

jbech-linaro:
I think there are two issues here
1) We stepped up to `v4.12` kernel for the `v2.6.0 release`, i.e, in the `xyz_stable.xml` we should have used `6e954e2f2cbd412f7bc874bb9145f69713194e52` instead of `50403184d40d04b3daf140417e031c16c2985eaf` in our stable xml's.

2) Repo cannot find tags thing: I guess that is true and indeed a limitation. But, I've said it before, I don't like that we just overwrite the `optee` branch without tagging (we've just recently started to tag the commit before doing the overwrite). By not having a "named" tag and just refering to the SHA-1, we are confusing the hxxl out of people and even the tool it self. For example if I go to our Linux tree and want to figure out where is the old tag located I run:
```bash
$ git branch -a --contains 50403184d40d04b3daf140417e031c16c2985eaf
```
And I get nothing. But still I can just do 
```bash
$ git checkout 50403184d40d04b3daf140417e031c16c2985eaf
```
That is **very** confusing and especially if you want to go back in history and try to figure out what has been going on. So, regardless if we are running repo, git submodules or any other tool I think we put/use named tag(s) where possible. I.e., basically in all gits we are controlling (OP-TEE and linaro-swg). For the ones where we are using upstream I guess we have to use either a SHA-1 or a relase tag from the upstream project and cross our fingers that they aren't overwriting their main branch (as we are doing).

vchong:
`Reviewed-by: Victor Chong <victor.chong@linaro.org>`

jbech-linaro:
Thanks @vchong !
Squashed, rebased and tag(s) applied (should be) ready for merge together with https://github.com/OP-TEE/build/pull/211! Thanks!

jbech-linaro:
Thanks @jforissier !

PR #90, opened 2018-01-29
OP-TEE 3.0.0 stable

jbech-linaro:
I let Travis build this ([results](https://travis-ci.org/jbech-linaro/manifest/builds/334664408)). Besides timing out (hitting 50 minutes) on the RPi3 build it seems to work fine. I've done a local test with repo for QEMU v7 and that worked fine (clone, build + xtest).

// Joakim

### DIFF
--- a/am43xx_stable.xml
+++ b/am43xx_stable.xml
@@ -5,15 +5,15 @@
   <remote fetch="https://github.com/OP-TEE" name="optee"/>
   <remote fetch="git://git.ti.com/ti-linux-kernel" name="ti-linux"/>
   <remote fetch="git://git.ti.com/ti-u-boot" name="ti-u-boot"/>
-  <project name="build.git" path="build" remote="optee" revision="refs/tags/2.6.0">
+  <project name="build.git" path="build" remote="optee" revision="refs/tags/3.0.0">
     <linkfile dest="build/Makefile" src="am43xx.mk"/>
   </project>
   <project name="busybox.git" path="busybox" remote="busybox" revision="dbf5a6da6a4295ce26edd1ce34fde567d19afa02"/>
-  <project name="gen_rootfs.git" path="gen_rootfs" remote="linaro-swg" revision="fe617d470e45778c909038bf3e7ca15174a4f893"/>
-  <project name="optee_client.git" path="optee_client" remote="optee" revision="refs/tags/2.6.0"/>
-  <project name="optee_examples.git" path="optee_examples" remote="linaro-swg" revision="refs/tags/2.6.0"/>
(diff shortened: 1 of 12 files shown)


jforissier:
- What about `am43xx_stable.xml`, `am57xx_stable.xml`? I suppose you're not updating them because we have no test report? In which case it sounds reasonable to me.
- What about `hikey960_stable.xml`?

jbech-linaro:
> What about am43xx_stable.xml, am57xx_stable.xml? I suppose you're not updating them because we have no test report? In which case it sounds reasonable to me.
>

Right.

> What about hikey960_stable.xml?

Bummer, that should also be included. I didn't have that in [my copy](https://github.com/jbech-linaro/github_scripts/blob/arch_linux/my_optee_admin_make_stable.sh#L2) of Pascal's script. I'll update the manifest for this platform manually, stay tuned!

igoropaniuk:
@jbech-linaro @jforissier I know that it's a bit late to do it now, but I can test AM57xx for 3.0.0. tag (I thought it was done, I didn't follow all comments about tested platforms. At least for me it would be convenient to have a stable manifest for 3.0.0), it will take 20 minutes.

The differences between DRA7 and AM57xx are minimal, so I don't think there would be any issues

jbech-linaro:
HiKey960 has been added and seems OK.

igoropaniuk:
AM57xx with 3.0.0 tag (without pager enabled)
```
+-----------------------------------------------------
35649 subtests of which 0 failed
68 test cases of which 0 failed
0 test case was skipped
TEE test application done!
```

jbech-linaro:
All xml's (stable) has been updated now.

igoropaniuk:
@jforissier 
> What about am43xx_stable.xml, am57xx_stable.xml? I suppose you're not updating them because we have no test report? In which case it sounds reasonable to me (but I don't think FVP or HiKey Debian have been tested either, have they?, so why update them by that logic).

Just in case, I've checked comments https://github.com/OP-TEE/optee_os/pull/2058, @glneo left R-b tag for all TI SoCs (`plat-ti`)

jforissier:
@igoropaniuk indeed, sorry, I guess I did not realize that `plat-ti` means `am43xx` and `am57xx`!

jbech-linaro:
@igor, @jforissier , it has been tested by @glneo , but still I'm a bit uneasy to add his T-B here, since although related, this is more about testing that things are working with the manifests being updated. I don't know what you are saying? Otherwise, I just need an A-B or R-B here so that I can squash it before merging.

jforissier:
@jbech-linaro I agree, IMO there is no reason to add any T-b here unless someone actually claims having tested this PR on a particular setup. For me this looks OK so:

jbech-linaro:

PR #91, opened 2018-02-16
Switch to GitHub for strace

jforissier:
SourceForge is unreliable [1]. Use GitHub instead.

Change-Id: I73241adf4a0ce9b0a8985914ee4523f1d46e9207
Link: [1] https://travis-ci.org/OP-TEE/build/jobs/342160257#L621-L629
Suggested-by: Joakim Bech <joakim.bech@linaro.org>

### DIFF
--- a/hikey.xml
+++ b/hikey.xml
@@ -6,12 +6,12 @@
 	<remote name="optee" fetch="https://github.com/OP-TEE" />
 	<remote name="savannah" fetch="git://git.savannah.gnu.org" />
 	<remote name="96b-hk" fetch="https://github.com/96boards-hikey" />
-	<remote name="sfnet" fetch="git://git.code.sf.net/p/strace" />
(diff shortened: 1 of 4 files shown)


jenswi-linaro:

jbech-linaro:

jforissier:
Note: fixed wrong URL in `qemu_v8.xml` (extra `/strace`).

jbech-linaro:
Not related to this PR and I'm not sure if repo will be used in the future, but since more and more gits are cloned directly from GitHub we could also in the future consider to reduce the amount of remotes to just use "github" as one remote replacing all individual GitHub remotes we are currently using. In that case you use a longer `name` instead (the full GitHub name). Then it would look something like I've done [here](https://github.com/jbech-linaro/manifest/blob/v8m/default.xml).

jforissier:
@jbech-linaro yep that would be good

PR #96, opened 2018-03-13
No upstream tracking

jbech-linaro:
Misc changes to all manifests and how we are working with them. Note, this is a manual copy/paste from the 3.0.0 tag, so it will probably take some iterations until everything builds and I'm letting Travis do the builds for me. So, therefore, ignore reviewing this until I think it is worth  reviewing.

Also, when this works, we need to update the instructions in build.git and a few other places (I will do that).

// Joakim


jbech-linaro:
@jforissier @vchong , I think this is ready for review now.

**Tests**:
- I have tested this on QEMU v7, QEMU v8, FVP and I've done a local build for HiKey 6220 to see that it compiles without errors. Having that said, I cannot run xtest on my HiKey, since my HiKey is being used in the HAB-setup and I no more HiKey's.
- Travis has built it "almost" successfully in [build#151](https://travis-ci.org/OP-TEE/manifest/builds/353743743). I cannot figure out why it says `QEMU v8` failed. I cannot see any issues in the log and after seeing that I did build it locally and that went fine (as well as running `xtest`).

**Changes**:
This patch set contains quite big and many changes.

1. **Stable release XML's are gone**. I'm going to update the instructions in build.git in a new PR telling how to get the old versions (and I have a new strategy for future releases).
2. Without asking anyone I have **deprecated the support for HiKey Debian and MTK8173**. 
3. I've **tweaked .travis.yml** so it has been cleaned up a bit and behaves a bit better in general. The ccache tweaks made no big difference, but I could see some improvements.
4. **Added guidelines and rules to a README.md**, so we have better control of how manifest XML's should be partitioned (this makes diff manifests much easier).
5. **Tagged all gits but the OP-TEE ones that are tracking `master`**. The initial patches in the series uses the `SHA1's` that was `HEAD` when I started this work. In later commits, I have replaced them with `refs/tags/foobar`. It was a pain in the .... to find those (and because of using tags, some gits are now using older commits than the revision used before this PR, but ... things still are working).
6. Changed **all manifest to mainly use GitHub as the default remote**. So now there are just one or two remotes in all xml-files.
7. **Added `clone-depth="1"` to all `refs/{tags/heads}/foobar`**, to save bandwidth and clone time.

// Joakim

.travis.yml
```diff
@@ -42,6 +41,8 @@ env:
   - $REPO_PROJ=rpi3
 
 before_script:
+  - export CCACHE_BASEDIR=$HOME/.ccache
```
jforissier: Does not sound good. `CCACHE_BASEDIR` is supposed to be a parent (prefix) of the build directory, so that more cache hits are observed when building the same tree under a different top-level directory.

Have you tried setting it to `$HOME/$REPO_PROJ`? That should increase cache hits by sharing files built for different platforms but with the same sources and compiler.

jbech-linaro:
Found the issue with QEMU v8 ([here](https://travis-ci.org/OP-TEE/manifest/jobs/353809205#L710-L713)), strange that I cannot see that when building locally, anyhow, now I know what to fix.

default.xml
```diff
@@ -13,14 +13,14 @@
         </project>
 
         <!-- linaro-swg gits -->
-        <project path="bios_qemu_tz_arm"   name="linaro-swg/bios_qemu_tz_arm.git" />
-        <project path="linux"              name="linaro-swg/linux.git" revision="optee"/>
```
jforissier: The problem with that is we're losing the branch information. So when someone wants to try out "the latest version", he does not know what to checkout. The solution to this may be to add a comment with the name of the branch.

[git submodules handle this much better, because you can specify the upstream branch to be used by `git submodule update --remote`, while the stored SHA1 will be used when `--remote` is not given].

jforissier:
@jbech-linaro maybe you have the libdtc dev package installed...

jbech-linaro:
> @jbech-linaro maybe you have the libdtc dev package installed...

Yes, I have that because of the apt-get dependencies in build.git and most distros have a device tree compile not from the stone age. So, I'm going to send another patch, one that removes it from the qemu_v8.xml and then we still get it for Travis using submodules in .travis.yml.

jforissier:
@jbech-linaro sounds reasonable. For "qemu v8: remove dtc as project":

.travis.yml
jbech-linaro: `CCACHE_BASEDIR`, you are right! This is wrong, I will fix. I should be either `$HOME` or probably even better as you said, `$HOME/$REPO_PROJ`.

default.xml
jbech-linaro: Yeah, this is what it is and initially I had exactly this in my commits, but I got rid of it later on since I was planning to add a comment in the documentation (either manifest or build.git, the former is probably better) about how to figure out the branch and tag corresponding to a commit. I think I prefer referring to that instead of cluttering down manifests with additional comments.
```bash
# Tag
$ git describe --tags `git rev-parse HEAD` --abbrev=0

# Branch
$ git branch -a --contains `git rev-parse HEAD`
```

jforissier: OK. `git rev-parse HEAD` is not needed, although you might want to keep just `HEAD` in the second line because otherwise it looks strange this `--contains` followed by nothing...

jbech-linaro: The `describe`-command is kind of misleading, since it gives you the tag closest to this. I.e, the commit is most likely **not** included in that tag. But the `branch`-command gives correct information.

jbech-linaro: > OK. git rev-parse HEAD is not needed

True, I used this yesterday when running a script, `HEAD` or just simply the `SHA1` in question is sufficient.

jforissier: HEAD is the default so it may be omitted

jbech-linaro: Right, but for the one looking at a manifest-xml, he will see a commit, so the `SHA1` must at least be mentioned. Anyhow, I'll add something about it as a documentation patch in a separate PR.

jbech-linaro:

PR #99, opened 2018-04-03
hikey: step up Arm-TF and atf-fastboot to match l-loader

jbech-linaro:
After pinning all git's but OP-TEE and at the same time introducing the new l-loader (using recovery.bin) we introduced a flash regression for HiKey. This also affected [HAB](https://github.com/jbech-linaro/hikey_auto_boot/tree/webserver_python3) which forced me to turn HAB off temporarily. The fix was pretty easy though. It turned out that we needed to step up both Arm-TF and atf-fastboot to make be in sync with l-loader.


jenswi-linaro:

jbech-linaro:

PR #101, opened 2018-04-05
Juno: Scp update

jenswi-linaro:


jenswi-linaro:
Depends on https://github.com/OP-TEE/build/pull/247

juno.xml
```diff
@@ -19,8 +19,8 @@
         <project path="optee_examples"       name="linaro-swg/optee_examples.git" />
 
         <!-- Misc gits -->
-        <project path="arm-trusted-firmware" name="ARM-software/arm-trusted-firmware.git" revision="refs/tags/v1.5-rc2" clone-depth="1" />
+        <project path="arm-trusted-firmware" name="ARM-software/arm-trusted-firmware.git" revision="refs/tags/v1.5" clone-depth="1" />
         <project path="busybox"              name="mirror/busybox.git"                    revision="refs/tags/1_24_0" clone-depth="1" />
         <project path="u-boot"               name="u-boot/u-boot.git"                     revision="refs/tags/v2018.03" clone-depth="1" />
-        <project path="vexpress-firmware"    name="arm/vexpress-firmware.git"             revision="refs/tags/armlt-20151204-002" clone-depth="1" remote="linaro"/>
+        <project path="vexpress-firmware"    name="arm/vexpress-firmware.git"             revision="670a8336738046ac910f4ed3746edc1b4ecf086c" clone-depth="1" remote="linaro"/>
```
jbech-linaro: With a pure commit you cannot use `clone-depth="1"`, see https://github.com/OP-TEE/manifest#26-when-to-use-clone-depth1 . So if there is a tag corresponding the commit, please use that instead of the pure comment (and then you can keep the `clone-depth`), otherwise if there is no tag, then please remove the clone-depth also.

jenswi-linaro:
Update

jbech-linaro:

PR #103, opened 2018-04-17
default, qemu_v8, hikey: update linux snapshot..

vchong:
..to include 94c2f2e (""tee: fix unbalanced context refcount in register
shm from fd") bug fix for 9f9806e ("tee: new ioctl to a register tee_shm
from a dmabuf file descriptor")

Change-Id: Id0f7f356597a451d20de21657b75b35597c980a0
Signed-off-by: Victor Chong <victor.chong@linaro.org>
Tested-by: Victor Chong <victor.chong@linaro.org> (qemu, qemu_v8, hikey)

### DIFF
--- a/default.xml
+++ b/default.xml
@@ -14,7 +14,7 @@
 
         <!-- linaro-swg gits -->
         <project path="bios_qemu_tz_arm"   name="linaro-swg/bios_qemu_tz_arm.git" revision="0271acbfa81d93e64c2fdd6ae8b0be14a8c45719" />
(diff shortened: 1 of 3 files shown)


jbech-linaro:
You're only including the ones that you have tested, right?

vchong:
> You're only including the ones that you have tested, right?

Yes. See T-b tag in commit. : )

Tag added.

Btw, PRs here don't trigger builds in HAB?

jbech-linaro:
> Btw, PRs here don't trigger builds in HAB?

No they don't, this is the only git where it's not really possible to use HAB, at least not as it is written right now.

PR #120, opened 2018-08-06
GCC 8

jforissier:


jforissier:
- Build-tested on all platforms with GCC 8.1 generated using the Jenkins server from the Linaro Toolchain group.
- Tested on QEMU (one error, see below), QEMUv8 (OK) and HiKey960 (OK).

Issues to be resolved before this can be merged:
1. `poplar.xml` won't build due to an outdated kernel (GCC 8 complains about `drivers/msp/hifb/src/drv_hifb_proc.c`)
2. `xtest 4005 5006` fails on QEMU (but strangely enough `xtest 5006` is OK):
  ```
  o regression_5006.3 Allocate_out_of_memory
    regression_5000.c:139: AllocateSharedMemory(&cs->context, &shm, SIZE_OVER_MEMORY_CAPACITY, 0x00000001) has an unexpected value: 0x0 = TEEC_SUCCESS, expected 0xffff000c = TEEC_ERROR_OUT_OF_MEMORY
    regression_5006.3 FAILED
  ```
3. Merge at the same time as https://github.com/OP-TEE/build/pull/286.

jforissier:
Ping

PR #122, opened 2018-11-06
Update QEMU, QEMUv8, HiKey620, HiKey960, Juno and FVP to kernel v4.19

jforissier:
HiKey960 is now properly supported by upstream kernel v4.19, so drop the
platform-specific fork and use the new tip of our multi-platform
linaro-swg/linux/optee branch (which was recently rebased onto v4.19).
While we're at it, also upgrade other platforms.

Change-Id: I77d03d71350808561dfff0c90a4a8307ce274d9a

### DIFF
--- a/default.xml
+++ b/default.xml
@@ -14,7 +14,7 @@
 
         <!-- linaro-swg gits -->
         <project path="u-boot"               name="linaro-swg/u-boot.git"                 revision="optee" />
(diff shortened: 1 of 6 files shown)


jforissier:
Tested on all platforms but Juno and HiKey (620) which I don't have.

jbech-linaro:
Looking OK.

PR #124, opened 2018-11-19
Add manifests to point to ms-iot forks

HernanGatta:
Create two new manifests for QEMU based of the ones for ARMv7 and ARMv8, respectively. These have the Linux, OP-TEE OS and OP-TEE Client repositories pointing to the corresponding forks under ms-iot. The rationale is that these forks have OCALL support, so that if developers want to test their OE enclaves under QEMU, they need build our forks instead of the upstream versions.

### DIFF
--- a/oe_default.xml
+++ b/oe_default.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<manifest>
+        <remote name="github" fetch="https://github.com" />
+
+        <default remote="github" revision="master" />
+
+        <!-- ms-iot gits -->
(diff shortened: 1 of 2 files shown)


HernanGatta:
Sorry, created this against the wrong base.

jenswi-linaro:
OCALL seems interesting, if I understand it correctly it's a way of doing RPC to the client process. Anything you'd like to upstream? I guess first step is to either discuss it in a new issue or at tee-dev@lists.linaro.org.

jforissier:
+1
Assuming it is indeed about client callbacks, then I've heard this requirement in the past and I'm sure several people could help implement/review/test such a feature upstream.

PR #127, opened 2018-12-17
hikey.xml: fix recover

jenswi-linaro:
Updates l-loader and atf-fastboot for make recovery boot work again.

Change-Id: Idcd83c9066773fc0e041c0d230bceacfb6ba7872
Signed-off-by: Jens Wiklander <jens.wiklander@linaro.org>


jbech-linaro:
Haven't tested it, but checked the commit in each tree and seems like this is latest in both of them, which is good. Thanks Jens!

jenswi-linaro:
Tag applied

PR #128, opened 2019-01-11
Tf a updates emmc fix

jbech-linaro:
This updates TF-A to include the patch the fixes the boot regression on the old HiKey. Unfortunately there is not tag to use, so I've change all xml-files to use a defined commit now.

I've tested this on HiKey
```bash
repo init -u https://github.com/jbech-linaro/manifest.git -b tf-a-updates-emmc-fix -m hikey.xml --reference ~/devel/reference
```

and QEMU-v8
```bash
repo init -u https://github.com/jbech-linaro/manifest.git -b tf-a-updates-emmc-fix -m qemu_v8.xml --reference ~/devel/optee_projects/reference
```

The same repo commands can be used if you want to give it a try on another platform who are using TF-A.

Both boot up as expected and xtest passes.

```bash
# Hikey
23665 subtests of which 0 failed
90 test cases of which 0 failed
0 test case was skipped
TEE test application done!
```

```bash
# QEMU v8
23299 subtests of which 0 failed
89 test cases of which 0 failed
0 test case was skipped
TEE test application done!
```

The second patch just sorts things alphabetically in accordance with the instructions in the README.md file.

### DIFF
--- a/default.xml
+++ b/default.xml
@@ -13,11 +13,11 @@
         </project>
 
         <!-- linaro-swg gits -->
-        <project path="u-boot"               name="linaro-swg/u-boot.git"                 revision="optee" />
         <project path="linux"                name="linaro-swg/linux.git"                  revision="221a1acee8047ae65c2d5980e3a7c5f73362c59d" />
         <project path="optee_benchmark"      name="linaro-swg/optee_benchmark.git" />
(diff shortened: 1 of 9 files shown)


jforissier:

PR #138, opened 2019-06-04
hikey960: update l-loader

jforissier:
Updates l-loader to the tip of branch testing/hikey960_v1.2. Fixes an
issue with "make recovery" remaining stuck at:

 NOTICE:  BL1-FWU: *******FWU Process Started*******

"git bisect" reports that the fix is commit 461b27f4900b ("hikey960:
add Trusted Board Boot feature"). This commit allows for a bigger size
for TF-A BL1. The problem was introduced when we upgraded TF-A from
v1.5-rc2 to v2.0 [1] but was unnoticed at the time.

Fixes: [1] 5af25841ef84 ("Update Arm TF-A to v2.0")
Change-Id: I3ec8d782c1e12e4f9b03fbf6b03c41f3729ccb9d


jenswi-linaro:
We clearly want this since Hikey960 is kind of crippled without it.

jbech-linaro:

PR #155, opened 2020-02-11
repo: remove symlink creation to gdb

jbech-linaro:
It seems like the newest version of repo (v2.0) no longer allows to
create symlinks to non-existent files. In the past repo has silently
ignored creating the symlink on the first repo sync. However the most
recent version gives an error instead.

Since our setup flow is to create toolchains after checking out the
repo forest, the only viable option we have is to no longer try to
create a symlink. This doesn't really affect anything as such, we just
had it there as a convenience so we could handle builds and debugging
from the same folder on disc.

### DIFF
--- a/default.xml
+++ b/default.xml
@@ -11,7 +11,6 @@
         <project path="optee_test"          name="OP-TEE/optee_test.git" />
         <project path="build"               name="OP-TEE/build.git">
                 <linkfile src="qemu.mk" dest="build/Makefile" />
-                <linkfile src="../toolchains/aarch32/bin/arm-linux-gnueabihf-gdb" dest="build/gdb" />
         </project>
 
         <!-- linaro-swg gits -->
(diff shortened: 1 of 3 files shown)


jforissier:

jbech-linaro:

a4lg:
I think it can be nice to have this change on other branches.

I am currently making a workaround by checking out older version of `repo` (tag `v1.13.10`) to sync older releases of OP-TEE (after normal sync of the `master` branch) like this:

```sh
cd $OPTEE_PROJECT/.repo/repo
git checkout v1.13.10
cd ../../
repo init -m qemu_v8.xml -b 3.8.0  # it is only an example
```

It works but... how about changing branches like `3.8.0`? It's not strictly necessary but helpful.

Appendix: If you want to revert this workaround and use default `repo`, run this:

```sh
( cd $OPTEE_PROJECT/.repo/repo; git checkout default )
```

jbech-linaro:
Even though this symlink stuff came from me initially and I've liked it and used it, I still think it's easier to just to:
```
../toolchains/aarch32/bin/arm-linux-gnueabihf-gdb
```
than checking out old versions etc. Another thing is that now since Python2 has been deprecated, you'll get into an issue with that since repo < v2 is Python2, but the most recent is written for Python3. So, no, I don't want to enable your proposal by default, but thanks for giving the suggestion.

a4lg:
Thanks for pointing it out.

PR #158, opened 2020-03-05
verdin-imx8mm: add initial support

igoropaniuk:
This introduces OP-TEE-based setup for Verdin i.MX8M Mini [1].
Here mainline vanilla TF-A/OP-TEE/U-Boot and Linux are used.
Currently using my github forks for Linux/U-Boot/TF-A as not all patches
(mostly fixes + Verdin DT hierarchy for Linux kernel) are being
mainlined to correspondent projects.

[1] https://www.toradex.com/de/computer-on-modules/verdin-arm-family/nxp-imx-8m-mini-nano
Signed-off-by: Igor Opaniuk <igor.opaniuk@gmail.com>

### DIFF
--- a/verdin.xml
+++ b/verdin.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<manifest>
+        <remote name="github"   fetch="https://github.com" />
+
+        <default remote="github" revision="master" />
+
+        <!-- OP-TEE gits -->
+        <project path="build"                name="OP-TEE/build.git" >
+                <linkfile src="verdin.mk" dest="build/Makefile" />
+        </project>
+        <project path="optee_client"         name="OP-TEE/optee_client.git" />
+        <project path="optee_os"             name="OP-TEE/optee_os.git" />
+        <project path="optee_test"           name="OP-TEE/optee_test.git" />
+
+        <!-- linaro-swg gits -->
+        <project path="optee_examples"       name="linaro-swg/optee_examples.git" />
+
+        <!-- temp igoropaniuk gits -->
+        <project path="linux"                name="igoropaniuk/linux"               revision="mainline_verdin" clone-depth="1" />
+        <project path="u-boot"               name="igoropaniuk/u-boot"              revision="verdin_latest" clone-depth="1" />
+        <project path="trusted-firmware-a"   name="igoropaniuk/trusted-firmware-a"  revision="verdin_latest" clone-depth="1" />
+
+        <!-- Misc gits -->
+        <project path="buildroot"            name="buildroot/buildroot.git"         revision="95942f5fcd35d783a49adce621ccf33480f1c88c" />
+</manifest>


verdin.xml
```diff
@@ -0,0 +1,25 @@
...
+        <project path="optee_os"             name="OP-TEE/optee_os.git" />
+        <project path="optee_test"           name="OP-TEE/optee_test.git" />
+
+        <!-- linaro-swg gits -->
+        <project path="optee_examples"       name="linaro-swg/optee_examples.git" />
+
+        <!-- temp igoropaniuk gits -->
+        <project path="linux"                name="igoropaniuk/linux"               revision="mainline_verdin" clone-depth="1" />
+        <project path="u-boot"               name="igoropaniuk/u-boot"              revision="verdin_latest" />
+        <project path="trusted-firmware-a"   name="igoropaniuk/trusted-firmware-a"  revision="verdin_latest" />
```
jforissier: Of course it would be better to have only official upstream repositories here, but I think we can accept this.
I'm not sure I understood your PR comment correctly: have you sent all the required patches upstream already or do you plan to do so?

igoropaniuk: @jforissier I do agree with you.

> I'm not sure I understood your PR comment correctly: have you sent all the required patches upstream already or do you plan to do so?

Some patches are sent/some not yet (like full hierarchy of device tree sources for Verdin module, as work is still in progress (video/usb nodes), which IMHO is not critical and has nothing in common with TEE-related stuff).

Just quick overview of custom changes:

**U-Boot** (https://github.com/igoropaniuk/u-boot/commits/verdin_latest)
U-Boot mainline master + 2 patches (in ML already) on top of it:
1. https://patchwork.ozlabs.org/patch/1249881/
2. https://patchwork.ozlabs.org/patch/1238011/ (not related to TF-A/OP-TEE, but prevents SPL to be booted properly)

**TF-A** (https://github.com/igoropaniuk/trusted-firmware-a/commits/verdin_latest):
1. UART fixes: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/3607
2. There is an issue, which occurs after secure state changing for the mem region, where BL32 resides `SET_SECURITY_STATE(bl32_image_ep_info.h.attr, SECURE)`(https://github.com/igoropaniuk/trusted-firmware-a/commit/34fd501f07723be30bf63e2ac13e30587822bb93) and leads to `BL32 initialization failed` (I haven't time to investigate this yet).

**Linux** (https://github.com/igoropaniuk/linux/commits/mainline_verdin).
1. Latest vanilla mainline + 1 patch on top of it which introduces two device tree sources on top it:
`arch/arm64/boot/dts/freescale/fsl-imx8mm-verdin.dtsi`
`arch/arm64/boot/dts/freescale/fsl-imx8mm-verdin-dev.dts`

As soon as changes are merged, I'll switch to the mainline repositories.
The final goal is to have quick shortcut to the setup, where only mainline TF-A/OP-TEE/U-Boot/Linux are used.

jforissier: OK, thanks for the details. This is fine with me.

verdin.xml
```diff
@@ -0,0 +1,25 @@
...
+        <project path="optee_client"         name="OP-TEE/optee_client.git" />
+        <project path="optee_os"             name="OP-TEE/optee_os.git" />
+        <project path="optee_test"           name="OP-TEE/optee_test.git" />
+
+        <!-- linaro-swg gits -->
+        <project path="optee_examples"       name="linaro-swg/optee_examples.git" />
+
+        <!-- temp igoropaniuk gits -->
+        <project path="linux"                name="igoropaniuk/linux"               revision="mainline_verdin" clone-depth="1" />
+        <project path="u-boot"               name="igoropaniuk/u-boot"              revision="verdin_latest" />
```
jbech-linaro:
```suggestion
        <project path="u-boot"               name="igoropaniuk/u-boot"              revision="verdin_latest" clone-depth="1" />
```
And same for the following, since it's named revision (see [this](https://optee.readthedocs.io/en/latest/building/gits/manifest.html#when-to-use-clone-depth-1)).

jbech-linaro: > And same for the following

And just above.

igoropaniuk:
Addressed Joakim's comment + applied Jerome's ack.

PR #159, opened 2020-04-03
Point to linaro-swg/linux.git branch optee

jforissier:
Instead of using a fixed commit for Linux, track the tip of the "optee"
Linaro branch, like we do for other Linaro Gits (optee_*).

Change-Id: Iac3572e2d0567f7007fa1f6d4404dc51b3747352

### DIFF
--- a/default.xml
+++ b/default.xml
@@ -14,7 +14,7 @@
         </project>
 
         <!-- linaro-swg gits -->
-        <project path="linux"                name="linaro-swg/linux.git"                  revision="1ad01d3482219ee7fcc5a4529f8f06570878bf72" />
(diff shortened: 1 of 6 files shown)


jenswi-linaro:

jbech-linaro:
I think we should have clone-depth set here to save some time when checking out code for CI etc.
https://optee.readthedocs.io/en/latest/building/gits/manifest.html#when-to-use-clone-depth-1

Other than that:

jforissier:
@jbech-linaro I wasn't sure if it was OK when revision is a branch name (as opposed to a "refs/tags/").
I'll add it.

PR #161, opened 2020-06-01
stm32mp1: initial support

etienne-lms:


etienne-lms:
It seems commit f4ffaa697d10d764444bfcfbe8791973e624245e was submitted to tag `3.9.0` but not to `master` branch.
@jforissier, could you force update `master` branch?

jenswi-linaro:
Why do we want f4ffaa6 on master branch?

etienne-lms:
Oh, sorry, I missed that tag were not pushed to master branch for this repo.
Discard my comment. I will update this P-R accordingly.

etienne-lms:
updated

stm32mp1-dev.xml
```diff
@@ -0,0 +1,27 @@
...
+        <project path="optee_client"        name="OP-TEE/optee_client.git" />
+        <project path="optee_docs"          name="OP-TEE/optee_docs.git" />
+        <project path="optee_os"            name="OP-TEE/optee_os.git" />
+        <project path="optee_test"          name="OP-TEE/optee_test.git" />
+        <project path="build"               name="OP-TEE/build.git" >
+                <linkfile src="stm32mp1.mk" dest="build/Makefile" />
+        </project>
+
+        <!-- linaro-swg gits -->
+        <project path="linux"                name="linaro-swg/linux.git" revision="optee" clone-depth="100" />
```
jbech-linaro, line 19: Any particular reason for clone-depth=100?

stm32mp1.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jbech-linaro, line 1: This file should be submitted to branch: https://github.com/OP-TEE/manifest/tree/3.9.0

stm32mp1.xml
```diff
@@ -0,0 +1,27 @@
...
+        <project path="optee_client"        name="OP-TEE/optee_client.git" revision="refs/tags/3.9.0" clone-depth="1" />
+        <project path="optee_docs"          name="OP-TEE/optee_docs.git" revision="refs/tags/3.9.0" clone-depth="1" />
+        <project path="optee_os"            name="OP-TEE/optee_os.git" revision="refs/tags/3.9.0" clone-depth="1" />
+        <project path="optee_test"          name="OP-TEE/optee_test.git" revision="refs/tags/3.9.0" clone-depth="1" />
+        <project path="build"               name="OP-TEE/build.git" revision="refs/tags/3.9.0" clone-depth="1" >
+                <linkfile src="stm32mp1.mk" dest="build/Makefile" />
+        </project>
+
+        <!-- linaro-swg gits -->
+        <project path="linux"                name="linaro-swg/linux.git" revision="5973cd5de37683006c42a570e38c2e41d0f70597" clone-depth="30" />
```
jbech-linaro, line 19: Cannot use clone-depth with a git commit as revision, see [here](https://optee.readthedocs.io/en/latest/building/gits/manifest.html#when-to-use-clone-depth-1).

stm32mp1.xml
jbech-linaro: Probably better to have the same name of the files, but on lives on the master branch and the other in the vX.Y.Z branch.

stm32mp1-dev.xml
etienne-lms: optee branch is a moving branch built on top of an official Linux tag.
I like the fact the fetched Linux repo shows the history from mainline tag to optee lastet commit.
I think 100 commits are enough to cover all evolution of optee branch.

stm32mp1.xml
etienne-lms: There are few diffs between the dev and stable refs: tf-a, u-boot and linux use the evolving branch in **stm32mp1-dev.xml** and immutable revision reference in **stm32mp1.xml**. Do you think it's an over complex way to release stm32mp1 support.

stm32mp1.xml
etienne-lms: Ok. Note however that this is used in all **\*.xml** file in tag X.Y.Z of the manifest repo, i.e. https://github.com/OP-TEE/manifest/blob/3.9.0/default.xml#L9.

stm32mp1-dev.xml
jbech-linaro: Fair enough.

stm32mp1.xml
jbech-linaro: You guys are obviously the main users of these manifests, so if this is what you need, then I'm fine with it. But the approach will differ to what we have otherwise for the rest of our setup's. Having said that, I'm kind of with you regarding the "dev", that it's nice to have a setup also tracking master non-OP-TEE gits, that makes it a breeze stepping up our dependents to a more recent version.

stm32mp1.xml
jbech-linaro: That is fine, that is a tag, there it works. It works with tags (x.y.z), branches (foobar), but not with SHA-1 (5a6b5af65ab65ab...).

jbech-linaro: Repo seems to have gotten an overhaul lately with the switch from Python2 to Python3, so it might be worth to see whether this still is a limitation or not.

etienne-lms: Ok. In the mean time i'll remove clone-depth info from this line.

jbech-linaro:

stm32mp1.xml
etienne-lms: In the end, I guess there is no reason for stm32mp1 platform to be handled differently than other platforms here. I can keep this stm32mp1-dev.xml in my own repo for my own convenience.
I'll remove this file.

Yet, I think the delivery script should be updated to not use "optee" branch a reference for Linux kernel repo but use the SHA1 of the commit that was used instead. But that's out of the scope of this P-R.

jbech-linaro: :+1: , overall I'm happy with this, do a final squash / force push.

jforissier: > Yet, I think the delivery script should be updated to not use "optee" branch a reference for Linux kernel repo but use the SHA1 of the commit that was used instead. But that's out of the scope of this P-R.

Some people want to have the latest version of everything, other people prefer stability.
Our current XMLs in the master branch are a compromise: they use the latest for "our stuff" (meaning all optee* repositories plus our linux kernel branch), while for all third-party projects they fix a version. This way we avoid regressions imported by external projects, while still allowing to work with the latest OP-TEE stuff. That's usually what we want as OP-TEE developers.
Stable XMLs are created for each release, and stored in the corresponding branch and tag (e.g., `3.9.0`). Such stable XMLs record SHA-1s for all projects.

We *could* create new XMLs tracking the latest for *all* projects, but honestly how hard it is to just pull a few projects manually when needed? (related question: is there a repo command  to make this easier, i.e., to checkout the tip of the tracking branch of a project that is pinned in the manifest? I think until recently it wasn't possible. Equivalent to `git submodule update` vs. `git submodule update --remote`).

etienne-lms: current way is fine.
The issue I see (1) in current way is the Linux kernel reference. It's not a SHA1, it's [`"optee"`](https://github.com/OP-TEE/manifest/blob/3.9.0/default.xml#L17). This one should also be a SHA1 I think.  

(1) the issue I **very lately see** :) it's used since quite a while now.

jforissier: Ah, you mean in the tagged manifest? Then yes I agree it's a mistake IMO. Sorry for the misunderstanding.

jforissier: `make_stable.sh` should record the SHA-1 IMO.

jbech-linaro: > We _could_ create new XMLs tracking the latest for _all_ projects, but honestly how hard it is to just pull a few projects manually when needed? 

If there is a regression when running other code together with OP-TEE we will see it regardless, so it's a matter finding, reporting and eventually fixing that right away or wait to a later point in time. Waiting is convenient, since our setup will be stable for longer periods of time. If it's a regression not caused or seen with introduction of OP-TEE, then it may just fly by ... unnoticed to us, which of course is good.

Waiting can also bite you in the end, since upgrading gits can be painful if you're waiting for too long. Even though it's annoying suddenly seeing a regression caused by another git that we usually doesn't touch, it's usually quick and easy to get it fixed. An email to the author of the patch causing the regression might be enough. "Hey, your patch caused a regression in setup X, Y ...", he'll likely try to address it right away, since it's all new.

Personally I wouldn't mind at all tracking master on all gits we are using for our bleeding edge development. We still tag and freeze it all every third month. So there is always a stable setup as a fallback if people don't want to use bleeding edge or having issues with it. With a bleeding edge setup, also the stable setup would run more up-to-date software otherwise, which might be beneficial. Take Buildroot for example, packages are constantly added and updated.

But, I defer this ... to someone else.

> (related question: is there a repo command to make this easier, i.e., to checkout the tip of the tracking branch of a project that is pinned in the manifest? I think until recently it wasn't possible. Equivalent to `git submodule update` vs. `git submodule update --remote`).

Don't know. Personally I use `repo forall -c` with different commands when doing bulk operations across the git's in the setup.

jbech-linaro: > `make_stable.sh` should record the SHA-1 IMO.

For stable, definitely yes. I think it used to be a stable until we just recently switch to track the branch (https://github.com/OP-TEE/manifest/commit/62f48b9a752c5ba91b1801250b713be8212eaf17). We missed that when making the release and therefore it too point to "optee".

etienne-lms:
Tag not applied yet.
Dependency on https://github.com/OP-TEE/build/pull/427.

etienne-lms:
tag applied

PR #162, opened 2020-06-02
qemu-v7,v8: step up QEMU version to 5.0.0

jbech-linaro:
Step up the QEMU version used by the OP-TEE builds for QEMU-v7 and -v8
to version 5.0.0. I've tested both of them and I've also tested running GDB (v7) and with `QEMU_VIRTFS_ENABLE=y` (v7). All seems to be working fine. The v8 writes this in the QEMU console
```
pflash_write: Write to buffer emulation is flawed
```
However, it doesn't seems to have any negative effects.

### DIFF
--- a/default.xml
+++ b/default.xml
@@ -22,6 +22,6 @@
 
         <!-- Misc gits -->
         <project path="buildroot"            name="buildroot/buildroot.git"               revision="95942f5fcd35d783a49adce621ccf33480f1c88c" />
-        <project path="qemu"                 name="qemu/qemu.git"                         revision="refs/tags/v3.1.0-rc3" clone-depth="1" />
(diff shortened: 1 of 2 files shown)


jforissier:
I have occasionaly run v5.0.0 too, I don't remember if it was v8 or v7, anyway it was working fine indeed.

jbech-linaro:

penpendandan:
I meet the same error. How to fix it?

jforissier:
> I meet the same error. How to fix it?

That question should probably be asked on the QEMU ML, it sounds like a QEMU bug to me.
https://git.qemu.org/?p=qemu.git;a=commit;h=4dbda935e054

PR #163, opened 2020-06-05
make_stable: use OP-TEE tag for linaro-swg linux repository

etienne-lms:
Change script to use OP-TEE release tag for linaro-swg github Linux
kernel repository when applicable to make sure stable release are
referenced with a immutable reference of the kernel source tree.

This change assumes OP-TEE release tag is also created for that
repository.

Signed-off-by: Etienne Carriere <etienne.carriere@linaro.org>

### DIFF
--- a/make_stable.sh
+++ b/make_stable.sh
@@ -53,5 +53,6 @@ do
 		sed "s/\(linaro-swg\/optee_examples.git\"\)/\1 revision=\"refs\/tags\/${VERSION}\" clone-depth=\"1\"/" |
 		sed "s/\(linaro-swg\/optee_benchmark.git\)revision.*/\1\/>/" | # Removes old revision
 		sed "s/\(linaro-swg\/optee_benchmark.git\"\)/\1 revision=\"refs\/tags\/${VERSION}\" clone-depth=\"1\"/" |
+		sed "s/\(linaro-swg\/linux.git\" *\)revision=\"optee\".*/\1revision=\"refs\/tags\/optee-${VERSION}\" clone-depth=\"1\"/" |
 		tee ${FILE} 2>&1 > /dev/null
 done


make_stable.sh
```diff
@@ -53,5 +53,7 @@ do
 		sed "s/\(linaro-swg\/optee_examples.git\"\)/\1 revision=\"refs\/tags\/${VERSION}\" clone-depth=\"1\"/" |
 		sed "s/\(linaro-swg\/optee_benchmark.git\)revision.*/\1\/>/" | # Removes old revision
 		sed "s/\(linaro-swg\/optee_benchmark.git\"\)/\1 revision=\"refs\/tags\/${VERSION}\" clone-depth=\"1\"/" |
+		sed "s/\(linaro-swg\/linux.git\)revision.*/\1\/>/" | # Removes old revision
+		sed "s/\(linaro-swg\/linux.git\"\)/\1 revision=\"refs\/tags\/${VERSION}\" clone-depth=\"1\"/" |
```
jforissier, line 57: That will work, but `${VERSION}` could be slightly misleading (Linux vs. OP-TEE version, and although Linux tags always start with a `v`). How about `optee-${VERSION}` instead?

etienne-lms: good point!
That's up to how OP-TEE tags its forked Linux. Dear admins, this is in your hands.
In the mean time let keep this P-R open and finalize it once all agreed.

jforissier: Also let's not forget to update https://optee.readthedocs.io/en/latest/general/releases.html#release-procedure once this is merged.

etienne-lms: From your funny face, do I understand that updating optee_docs as you suggest is sufficient to decide how OP-TEE Linux kernel tree is tagged?
If so, then i'll create a P-R in optee_docs...

jforissier: No, let's decide here and we can update this PR if needed and the doc independently.

@jbech-linaro, do you prefer the simplest option (`${VERSION}`) or the slightly more complex/inconsistent but slightly less misleading one (`optee-${VERSION}`)? I don't have a strong opinion either way.

jbech-linaro: Same here, no strong opinion, but since there already are tags in the kernel tree, I guess your proposal is better (for the Linux kernel fork git only).

etienne-lms: +1 for `optee-X.Y.Z` tagging when using an external repository.

> (for the Linux kernel fork git only).

I think this applies to any external repo that has been forked in OP-TEE project, as U-Boot for Qemu/Armv7 support that currently does use a local `optee` branch (Actually, it did, since Qemu can move ahead to mainline, as done in #164).

Maybe current proposal should be update to cover all lines in the manifest file that match both `linaro-swg` github account and `optee` as revision reference, not only the Linux kernel source tree.

jforissier: > Maybe current proposal should be update to cover all lines in the manifest file that match both `linaro-swg` github account and `optee` as revision reference, not only the Linux kernel source tree.

We need linux only for now, let's stick to that. Even if your proposal sounds logical, things can happen that would make it insufficient or wrong. There is no need to anticipate IMO.

etienne-lms:
Removed [RFC] from P-R title.
Updated with a review fixup commit addressing review comments.

jforissier:

jforissier:
Ping?
And https://github.com/OP-TEE/optee_docs/pull/80.

etienne-lms:
fixup commit squashed and tag applied.
@jbech-linaro, your r-b tag is still applicable?

jbech-linaro:
> @jbech-linaro, your r-b tag is still applicable?

yes!

etienne-lms:
tags applied.

PR #167, opened 2020-08-06
qemu: remove deprecated linkfile elements with .. in the path

jforissier:
Newer versions of repo can't build older OP-TEE branches due to the
following error:

 error.ManifestInvalidPathError: <linkfile> invalid "src": ../toolchains/aarch32/bin/arm-linux-gnueabihf-gdb: bad component: ..

This commit removes the problematic elements in the QEMU manifests. They
were just there for convenience.

Change-Id: If94077ad7c6336e38f0fbcbfe3961037ba6f41b9


jforissier:
Note this is against branch 3.8.0, not master.

jforissier:
Sorry I should cherry-pick the commit from master instead.

PR #168, opened 2020-08-06
repo: remove symlink creation to gdb

jforissier:
It seems like the newest version of repo (v2.0) no longer allows to
create symlinks to non-existent files. In the past repo has silently
ignored creating the symlink on the first repo sync. However the most
recent version gives an error instead.

Since our setup flow is to create toolchains after checking out the
repo forest, the only viable option we have is to no longer try to
create a symlink. This doesn't really affect anything as such, we just
had it there as a convenience so we could handle builds and debugging
from the same folder on disc.

Acked-by: Jerome Forissier <jerome@forissier.org>
(cherry picked from commit 017ade90e840a9da45c775e7146f322889986ca7)
Change-Id: Ief8cb22f7b2d8c9a2f88e9e817306883f6c02f5b


jbech-linaro:

PR #171, opened 2020-09-21
Update buildroot to 2020.08

jforissier:
Supports building with GCC 9.

Change-Id: Ieaa3a304ef3e414d0d067072d36e82a2eb2cc0e8

### DIFF
--- a/am43xx.xml
+++ b/am43xx.xml
@@ -17,7 +17,7 @@
         <project path="optee_examples" name="linaro-swg/optee_examples.git" />
 
         <!-- Misc gits -->
-        <project path="buildroot"      name="buildroot/buildroot.git"             revision="95942f5fcd35d783a49adce621ccf33480f1c88c" />
+        <project path="buildroot"      name="buildroot/buildroot.git"             revision="refs/tags/2020.08" clone-depth="1" />
         <project path="linux"          name="ti-linux-kernel/ti-linux-kernel.git" revision="00a3beacce33dc14fa301eb5f3fb5a341212e9b4" remote="ti" />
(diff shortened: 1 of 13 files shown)


jenswi-linaro:

etienne-lms:
Hi all. Despite I expected no issues with such upgrades, testing on stm32mp1 pointed at least an issue in the partition table of the generated image. Seems not related to gcc9.2 even if I doubt the issue is related buildroot. I'll investigate. Could you please wait a bit before merging this change and https://github.com/OP-TEE/build/pull/438?

jforissier:
@etienne-lms sure. Since this (and the GCC 9.2 update) affects all platforms, I'm happy to wait until this is tested well enough.

etienne-lms:
Thanks @jforissier. Actually I've seen a weird thing in my setup. Buildroot was pointing to its master, not tag 2020.08. I'm rebuilding the whole thing and I am pretty confident there is no real issue. I'll tell you by then.

etienne-lms:
Issue is indeed in the partition table of the generated image (stm32mp1 builds a raw SDcard image with a GPT partitioning). The issue comes from Buildroot 2020.08 has upgraded to genimage v13 which creates a hybrid MBR partition table (https://github.com/pengutronix/genimage/pull/96) for stm32mp1 platform. TF-A boot stage does not support hybrid MBR. After few investigation, this root cause is stm32mp1 partitioning description. I've created https://github.com/OP-TEE/build/pull/439 to fix the issue. Once merged, buildroot can be upgraded without issues regarding for stm32mp1.

etienne-lms:
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`

Issue #179, opened 2021-03-16
OPTEE build failure due to buildroot's incompatibility with glibc 2.33


xiangmy:
Dear OPTEE group,

The glibc `2.33` removed `_STAT_VER` and the upstream buildroot has reflected this change (https://git.busybox.net/buildroot/commit/?id=f45925a951318e9e53bead80b363e004301adc6f). So, I'm wondering whether OPTEE wants to update the buildroot to `2021.02`.

PS: I've tested the build process with `qemu_v8.xml` on my branch: [89c5402](https://github.com/xiangmy/manifest/commit/89c54026d3998bee56ef462dcc8ff0ec2b22992b). I also attach the [log](https://github.com/OP-TEE/manifest/files/6145985/log.txt) of my failed build.

BTW, the edk2 is also incompatible with `python 3.9`. For instance, `python 3.9` removed `tostring()` and `fromstring`, and this will cause errors while building.

Thanks,
Mingyuan

jforissier:
Hi @xiangmy,

Have you worked on upgrading Buildroot and EDK2 already? Or have you just met these errors? If you have patches to propose please do and I will be glad to review them. Otherwise I can work on the upgrade.

xiangmy:
Hi @jforissier ,

I can clean up my changes and push them.

Thanks,
Mingyuan

jforissier:
@xiangmy that would be very helpful, thanks!

xiangmy:
Hi @jforissier ,

Sorry for the late update. Upgrading the EDK2 is quite painful. I tried various of commits/tags, but they all failed at some warnings with `cc1: all warnings being treated as errors`, then the make process failed. (Here is an example of a failed log: [log1.txt](https://github.com/OP-TEE/manifest/files/6155045/log1.txt)) However, I manage to compile the whole system with only the following patches applied, which fix the python 3.9 compatibility:
+ [5d8648345cd9ad729837118c731063c59edea192](https://github.com/tianocore/edk2/commit/5d8648345cd9ad729837118c731063c59edea192)
+ [43bec9ea3d56f3662ede78023baa2a791b66acac](https://github.com/tianocore/edk2/commit/43bec9ea3d56f3662ede78023baa2a791b66acac)
+ [20b292d0cdf7dce58d824fdf9ab1613c2a1ad2ec](https://github.com/tianocore/edk2/commit/20b292d0cdf7dce58d824fdf9ab1613c2a1ad2ec)

[...]

PR #180, opened 2021-03-18
Update Buildroot, EDK2 and QEMU

xiangmy:
As disscussed in https://github.com/OP-TEE/manifest/issues/179
+ Buildroot fixed the glibc 2.33 compatibility [issue](https://git.busybox.net/buildroot/commit/?id=f45925a951318e9e53bead80b363e004301adc6f) (`_STAT_VER` was removed in glibc 2.33), we should upgrade the Buildroot to reflect that change.
+ EDK2 fixed its incompatibility with python 3.9, so we should update it to edk2-stable202102
+ QEMU fails to build with binutils 2.35 (https://bugs.launchpad.net/ubuntu/+source/binutils/+bug/1907789), hence, we need to update it to the latest version.

### DIFF
--- a/default.xml
+++ b/default.xml
@@ -21,7 +21,7 @@
         <project path="soc_term"             name="linaro-swg/soc_term.git" />
 
         <!-- Misc gits -->
-        <project path="buildroot"            name="buildroot/buildroot.git"               revision="refs/tags/2020.08" clone-depth="1" />
+        <project path="buildroot"            name="buildroot/buildroot.git"               revision="refs/tags/2021.02" clone-depth="1" />
         <project path="qemu"                 name="qemu/qemu.git"                         revision="refs/tags/v5.1.0" clone-depth="1" />
         <project path="trusted-firmware-a"   name="TF-A/trusted-firmware-a.git"           revision="refs/tags/v2.3" remote="tfo" />
         <project path="u-boot"               name="u-boot.git"                            revision="refs/tags/v2020.04" remote="u-boot" clone-depth="1" />
--- a/default.xml
+++ b/default.xml
@@ -22,7 +22,7 @@
 
         <!-- Misc gits -->
         <project path="buildroot"            name="buildroot/buildroot.git"               revision="refs/tags/2021.02" clone-depth="1" />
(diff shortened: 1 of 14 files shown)


default.xml
```diff
@@ -21,8 +21,8 @@
         <project path="soc_term"             name="linaro-swg/soc_term.git" />
 
         <!-- Misc gits -->
-        <project path="buildroot"            name="buildroot/buildroot.git"               revision="refs/tags/2020.08" clone-depth="1" />
-        <project path="qemu"                 name="qemu/qemu.git"                         revision="refs/tags/v5.1.0" clone-depth="1" />
+        <project path="buildroot"            name="buildroot/buildroot.git"               revision="refs/tags/2021.02" clone-depth="1" />
+        <project path="qemu"                 name="qemu/qemu.git"                         revision="39912c14da07a2dbc73854addcfa0a42596340ac" clone-depth="1" />
```
jforissier, line 25: - Please cite the QEMU commit that fixes the issue in the commit descriptions: `https://github.com/qemu/qemu/commit/bbd2d5a812`
- I'd prefer a tag here but it looks like the above commit has not reached a release yet? If so it is OK to keep a SHA1 revision.

jforissier (changes requested):
Please split into 3 commits, one for each component you are upgrading.
And see my comment below. Thanks!

default.xml
xiangmy: > I'd prefer a tag here but it looks like the above commit has not reached a release yet? If so it is OK to keep a SHA1 revision.

Yes, I cannot find a stable version that covers this change. I've updated the commit message, and it will look better now.

jforissier:
- For commit "Update Buildroot":
Please specify the version in the commit subject: "Update Buildroot to version 2021.02". With that:
- For commit "Update EDK2":
Please add prefix: "qemu_v8: update EDK2". With that:
- For commit "Update QEMU":
Please add prefix: "qemu, qemu_v8: update QEMU". With that:

Mainline EDK2 is also used by `fvp.xml`, perhaps someone who can test it can try to upgrade?

jforissier:
Small nit: no blank lines should be added between tags (Link:, Signed-off-by:, Acked-by: etc.).

xiangmy:
Got you! Thanks for the reviewing!

jforissier (approved):

PR #200, opened 2021-09-02
add rust sdk entry to qemu_v8.xml

fanrong1992:
Signed-off-by: Rong Fan <fanrong03@baidu.com>
Yuan Zhuang <zhuangyuan04@baidu.com>

### DIFF
--- a/qemu_v8.xml
+++ b/qemu_v8.xml
@@ -27,4 +27,5 @@
         <project path="trusted-firmware-a"   name="TF-A/trusted-firmware-a.git"           revision="refs/tags/v2.5" clone-depth="1" remote="tfo" />
         <project path="u-boot"               name="u-boot.git"                            revision="refs/tags/v2021.04" remote="u-boot" clone-depth="1" />
         <project path="xen"                  name="xen.git"                               revision="refs/tags/RELEASE-4.14.1" remote="xen-git" clone-depth="1" />
+        <project path="optee_rust"           name="apache/incubator-teaclave-trustzone-sdk.git"            revision="dfd51b7286407070e40b23defe6cef33f4f120c7" />
 </manifest>


qemu_v8.xml
```diff
@@ -19,6 +19,9 @@
         <project path="optee_benchmark"      name="linaro-swg/optee_benchmark.git"/>
         <project path="optee_examples"       name="linaro-swg/optee_examples.git" />
 
+        <!-- apache gits-->
+        <project path="optee_rust"          name="apache/incubator-teaclave-trustzone-sdk"                  clone-depth="1" />
```
jbech-linaro, line 23: Nit picking comment: `name` isn't aligned with the other `name` in the file. It'd be great of you could fix that.

jbech-linaro (commented):
Just want to mention the documentation page where we talk a bit about how we structure the manifest files and some "rules". Besides the "nit", this patch seems good, but in case you have seen the  documentation page, it could be good to know that it exists (https://optee.readthedocs.io/en/latest/building/gits/manifest.html)

jenswi-linaro:
Please use one  `Signed-off-by` per name, with your own at the bottom.

qemu_v8.xml
```diff
@@ -19,6 +19,9 @@
         <project path="optee_benchmark"      name="linaro-swg/optee_benchmark.git"/>
         <project path="optee_examples"       name="linaro-swg/optee_examples.git" />
 
+        <!-- apache gits-->
+        <project path="optee_rust"           name="apache/incubator-teaclave-trustzone-sdk"                  clone-depth="1" />
```
jenswi-linaro, line 23: Does this git include the Rust examples?

mssun: Yes, the Rust examples are in the `examples` directory in this repo.

jenswi-linaro:

qemu_v8.xml
jenswi-linaro: Thanks

jbech-linaro:

ruchi393:
`Acked-by: Ruchika Gupta <ruchika.gupta@linaro.org>`

qemu_v8.xml
```diff
@@ -19,6 +19,9 @@
         <project path="optee_benchmark"      name="linaro-swg/optee_benchmark.git"/>
         <project path="optee_examples"       name="linaro-swg/optee_examples.git" />
 
+        <!-- apache gits-->
```
etienne-lms, line 22: bitpicking: space char after gits

etienne-lms (commented):
Do we expect to pick the latest master of teaclave-trustzone, or some known/stable commit/tag?

mssun:
> Do we expect to pick the latest master of teaclave-trustzone, or some known/stable commit/tag?

We plan to put the latest mater first, then pin to a revision (stable tag or commit) when releasing a new OP-TEE version each time. Similar with the other OP-TEE projects in the manifest file.

jforissier:
Author doesn't match `Signed-off-by`.

qemu_v8.xml
```diff
@@ -19,6 +19,9 @@
         <project path="optee_benchmark"      name="linaro-swg/optee_benchmark.git"/>
         <project path="optee_examples"       name="linaro-swg/optee_examples.git" />
 
+        <!-- apache gits -->
+        <project path="optee_rust"           name="apache/incubator-teaclave-trustzone-sdk.git"                  clone-depth="1" />
```
jforissier, line 23: You need `sync-s="true", don't you?

jforissier:
The `optee_rust` directory is 1.4 GB after cloning is complete. I am really not sure we want to impose that on all users given than most won't use Rust. I think this should be a separate manifest.

jbech-linaro:
> The `optee_rust` directory is 1.4 GB after cloning is complete. I am really not sure we want to impose that on all users given than most won't use Rust. I think this should be a separate manifest.

Is that after repo sync? Or repo sync and then git submodule update? If it's the former, then I think I'm with you. But if the repo sync size and time difference is negligible and you add a "git submodule ..." to the docs explaining how to build optee_rust, then I think it could be OK.

Having said that, I'm not really opposed to a separate manifest either, that could also work.

jforissier:
> > The `optee_rust` directory is 1.4 GB after cloning is complete. I am really not sure we want to impose that on all users given than most won't use Rust. I think this should be a separate manifest.
> 
> Is that after repo sync? Or repo sync and then git submodule update? 

That is after repo sync + git submodule update (or just after repo sync if sync-s="true" of course).

> If it's the former, then I think I'm with you. But if the repo sync size and time difference is negligible and you add a "git submodule ..." to the docs explaining how to build optee_rust, then I think it could be OK.

Yeah "git submodule update" seems to be enough. We could live with that especially since the majority of people won't need it.
 
> Having said that, I'm not really opposed to a separate manifest either, that could also work.

vchong:
Another option can be to use the `groups` option to `repo init`.
```
<project path="optee_rust" name="apache/incubator-teaclave-trustzone-sdk.git" clone-depth="1" groups="notdefault,rust" />
```

Then a 'normal' `repo init -u https://github.com/OP-TEE/manifest.git -m qemu_v8.xml; repo sync` will not pick up `optee_rust`; only `repo init -u https://github.com/OP-TEE/manifest.git -m qemu_v8.xml -g "default,notdefault,rust"; repo sync` will.

Extending this further, we can even have a `common.xml` and have each `platform.xml` include it instead of repeatedly specifying the same repo in each `platform.xml`. I've not really tried this myself though; just seen manifests in other projects that seem to make this possible.

jbech-linaro:
> Another option can be to use the `groups` option to `repo init`.

I've learnt something today, I wasn't aware of this feature, thanks for sharing it (and yes, that would be an alternative as well).

qemu_v8.xml
fanrong1992: `sync-s="true"` is not needed. The default `optee_rust` directory is 3.6M. If someone need to use Rust, `setup.sh` in `optee_rust` could setup Rust environment and initialize submodules. The `optee_rust` directory is 602M after initializing is complete. The building and running details will be given as an `optee_docs` commit today.

mssun:
> `sync-s="true"` is not needed. The default `optee_rust` directory is 3.6M. If someone need to use Rust, `setup.sh` in `optee_rust` could setup Rust environment and initialize submodules. The `optee_rust` directory is 602M after initializing is complete. The building and running details will be given as an `optee_docs` commit today.

As @fanrong1992 said, we don't have to clone all submodules recursively. In the Teaclave TrustZone SDK repo, we have a `setup.sh` script to 1) only initialize necessary submodules and 2) install correct version of Rust toolchains.

For people who want to build Rust examples, they need to run `setup.sh` before building OP-TEE projects, which is not automated in the `build` (https://github.com/OP-TEE/build/pull/521) system of OP-TEE.

Of course, this will not be a problem when we can upstream the OP-TEE target to the Rust compiler.

jforissier:
> > `sync-s="true"` is not needed. The default `optee_rust` directory is 3.6M. If someone need to use Rust, `setup.sh` in `optee_rust` could setup Rust environment and initialize submodules. The `optee_rust` directory is 602M after initializing is complete. The building and running details will be given as an `optee_docs` commit today.
> 
> As @fanrong1992 said, we don't have to clone all submodules recursively. In the Teaclave TrustZone SDK repo, we have a `setup.sh` script to 1) only initialize necessary submodules and 2) install correct version of Rust toolchains.
> 
> For people who want to build Rust examples, they need to run `setup.sh` before building OP-TEE projects, which is not automated in the `build` ([OP-TEE/build#521](https://github.com/OP-TEE/build/pull/521)) system of OP-TEE.
> 
> Of course, this will not be a problem when we can upstream the OP-TEE target to the Rust compiler.

OK, sounds good. Still, I would like complete instructions in the commit text of build.git PR521 so that I (or anyone else) can reproduce the build easily.

I am fine with this PR then. The subject line had better be "qemu_v8: add Rust SDK entry" for consistency with the format we use in general.

qemu_v8.xml
jforissier: One more thing (thanks @etienne-lms :wink:): this brings a risk of breaking as upstream changes, which is out of our control. Like other dependencies (Buildroot,EDK2, TF-A, etc.) this should be a fixed revision.

jbech-linaro: I'm probably the one to blame here, in the call with Mingshen I explained that we have two ways to track things.
1) The (release)tagged manifest XML's contains only tagged versions.
2) The non-tagged manifest XML's point to master on the OP-TEE gits and the rest tagged. This work is about optee_rust, so I thought we should treat that the same as the rest of the OP-TEE gits.

jforissier: @jbech-linaro thanks for the clarification. IMO we can't really consider optee_rust is similar to the other optee_* repositories because we OP-TEE maintainers don't have any control on it. So for example we can't synchronize updates or add quick fixes to master like we do for instance with optee_os / optee_client /  optee_test should something break.

That being said, this is not a strong opinion. We could as well track master and only if things are unstable we would then switch to pinned versions. I guess it all boils down to the way optee_rust is developed: @fanrong1992 can we expect optee_rust master to be reasonably stable (like optee_os)?

mssun: I think @jforissier raised a valid concern. Teaclave TrustZone SDK (optee_rust) is still in actively development, which all changes are merged into the master branch.

Right now, we have an active PR on integrating to the build system (https://github.com/OP-TEE/build/pull/521). This may also introduce new changes. We can put a commit hash in the revision here once that PR is finalized.

For a stable release tag of Teaclave TrustZone SDK, it may take some time to publish a new release with the recent changes of the integration. We'll come up new PR when a new stable release published from Teaclave.

etienne-lms: > For a stable release tag of Teaclave TrustZone SDK, it may take some time to publish (...)

I think a commit sha1 is fine as reference, we don't really need a stable reference tag from teaclave for now.

qemu_v8.xml
```diff
@@ -19,6 +19,9 @@
         <project path="optee_benchmark"      name="linaro-swg/optee_benchmark.git"/>
         <project path="optee_examples"       name="linaro-swg/optee_examples.git" />
 
+        <!-- apache gits -->
+        <project path="optee_rust"           name="apache/incubator-teaclave-trustzone-sdk.git"              revision="76680044fbd7e0c5260474aa6acd4d901258842b" clone-depth="1" />
```
jforissier, line 23: One more thing: who else thinks this had better go with the other gits in the `<!-- Misc gits -->` section?

jenswi-linaro: Makes sense to me.

jbech-linaro: No strong opinion, but I was thinking the same when I read this patch the first time.

qemu_v8.xml
```diff
@@ -27,4 +27,5 @@
         <project path="trusted-firmware-a"   name="TF-A/trusted-firmware-a.git"           revision="refs/tags/v2.5" clone-depth="1" remote="tfo" />
         <project path="u-boot"               name="u-boot.git"                            revision="refs/tags/v2021.04" remote="u-boot" clone-depth="1" />
         <project path="xen"                  name="xen.git"                               revision="refs/tags/RELEASE-4.14.1" remote="xen-git" clone-depth="1" />
+        <project path="optee_rust"           name="apache/incubator-teaclave-trustzone-sdk.git"            revision="76680044fbd7e0c5260474aa6acd4d901258842b" clone-depth="1" />
```
jbech-linaro, line 30: Thanks for adding the revision, but when using a `sha1` and not a tag, then `clone-depth` doesn't work, see the OP-TEE documentation covering the manifests. So, in short I'm asking you to keep the `sha1`, but remove the `clone-depth`.
https://optee.readthedocs.io/en/latest/building/gits/manifest.html#when-to-use-clone-depth-1

jforissier: @jbech-linaro when I tried, it mostly worked although I got a scary message from `git log`... (below `git lol` is an alias for `git log --prety=<some cool format>`)
```
[...]
repo sync has finished successfully.
jerome@matebook:~/work/optee_repo_qemu_v8$ cd optee_rust/
jerome@matebook:~/work/optee_repo_qemu_v8/optee_rust ((7668004...))$ git lol
7668004 4 days ago BruceFan add setup.sh to setup rust environment and initialize rust submodule (#40)error: Could not read 27af4190228023e296464994a533f9572feec505
fatal: Failed to traverse parents of commit 03cfbb245cdd16a65f2ea5884ed4830e56f2e88c
```

jbech-linaro: > when I tried, it mostly worked 


OK, Google might have "improved" repo, because I'm pretty sure it failed hard in the past.

> although I got a scary message from `git log`...

Yeah, I think I'd prefer having a clean looking log when checking out things, to avoid questions and confusion.

fanrong1992: > Thanks for adding the revision, but when using a `sha1` and not a tag, then `clone-depth` doesn't work, see the OP-TEE documentation covering the manifests. So, in short I'm asking you to keep the `sha1`, but remove the `clone-depth`.
> https://optee.readthedocs.io/en/latest/building/gits/manifest.html#when-to-use-clone-depth-1

OK, `clone-depth` has been removed.

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@linaro.org>`

jbech-linaro:

jforissier (approved):

jforissier:
Merged manually. I moved the `optee_rust` line in alphabetical order and removed the dupplicate `Reviewed-by`.

PR #203, opened 2021-09-29
Update buildroot to 2021.08

jforissier:
Upgrading to the latest Buildroot release (2021.08).

Change-Id: I28f1fdebf4e68cb0833a352143c63a668c6a9250

### DIFF
--- a/am43xx.xml
+++ b/am43xx.xml
@@ -17,7 +17,7 @@
         <project path="optee_examples" name="linaro-swg/optee_examples.git" />
 
         <!-- Misc gits -->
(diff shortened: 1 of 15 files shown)


jbech-linaro:
Wait merge until after `3.15.0` or?

jforissier:
> Wait merge until after `3.15.0` or?

Either way is fine with me, this is just an update to stay current, but I think it's safe to take it now. We can easily revert any platform should something wrong happen when testing for the release.

jbech-linaro:
> Either way is fine with me, this is just an update to stay current, but I think it's safe to take it now. We can easily revert any platform should something wrong happen when testing for the release.

Sure, I'll let you decide.

PR #208, opened 2022-01-10
Upgrade trusted-firmare-a to v2.6

jforissier:
Upgrade all platforms to the latest v2.x version of TF-A: QEMU, QEMUv8,
FVP, HiKey, HiKey960, Juno, RPi3 and STM32MP1.

Change-Id: I536be2a27d03569ad09467022ac80c11848e6feb


jenswi-linaro:

jforissier:
Oops, I realize I updated only the platform that were already using the latest version (v2.5) and ignored `imx.xml` (at v2.3), `poplar.xml` and `synquacer.xml` (at v2.2). Should I update them too? I think it would be preferable since I see no reason not too.

jenswi-linaro:
Please do, this is a good time to get everything tested.

jenswi-linaro:
Looks good

PR #215, opened 2022-06-13
fvp: add Trusted Services support

balint-dobszay-arm:
Add new manifest file derived from the existing fvp.xml which includes
the Trusted Services project and its dependencies.

Related PR: https://github.com/OP-TEE/build/pull/571

### DIFF
--- a/fvp-ts.xml
+++ b/fvp-ts.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<manifest>
+        <include name="fvp.xml" />
+
+        <remote name="arm-gitlab" fetch="https://git.gitlab.arm.com" />
+
+        <!-- OP-TEE gits -->
+        <!-- Need to remove and re-add to replace Makefile symlink -->
+        <remove-project path="build"                    name="OP-TEE/build.git" />
+        <project        path="build"                    name="OP-TEE/build.git">
+                <linkfile src="fvp-psa-sp.mk" dest="build/Makefile" />
+        </project>
+
+        <!-- linaro-swg gits -->
+        <!-- Replace Linux with mainline version -->
+        <remove-project path="linux"                    name="linaro-swg/linux.git" />
+        <project        path="linux"                    name="torvalds/linux.git"                       revision="refs/tags/v5.16" clone-depth="1" />
+
+        <!-- Misc gits -->
+        <!-- The fTPM is not used in this config -->
+        <remove-project path="ms-tpm-20-ref"            name="microsoft/ms-tpm-20-ref" />
+        <!-- Add Trusted Services Linux drivers -->
+        <project        path="linux-arm-ffa-user"       name="linux-arm/linux-trusted-services.git"     revision="refs/tags/v4.0.0" clone-depth="1" remote="arm-gitlab" />
+        <project        path="linux-arm-ffa-tee"        name="linux-arm/linux-trusted-services.git"     revision="refs/tags/tee-v1" clone-depth="1" remote="arm-gitlab" />
+        <!-- Add Trusted Services project -->
+        <project        path="trusted-services"         name="TS/trusted-services.git"                  revision="integration" remote="tfo" />
+</manifest>


gyuri-szing:
> What is the purpose with this?

Good question, something we might have forgotten to ask ourselves.
Since our team is responsible to develop the OP-TEE SPMC and PSA SPs, we started using the OP-TEE SW integration methodology, which is to use this repo. What we try to do here is to eliminate our fork of this repository and up-stream what we have.
This has two main components:
- having an integration where all components are configured right to enable using OP-TEE SPMC. The scope of this can vary. As a minimum this means TF-A and OP-TEE configuration, but that is a partial implementation as you won't have any SPs to run and Linux configuration is missing too. The other extreme is what we do here, to add TS to the integration and configure everything as needed.
- enabling SPMC testing with TS SPs and verify that in OP-TEE CI. This is out of scope for this PR, but is not possible without it.

Taking a step back, what we try to do here is to make the PSA SPs a first class citizen of the OP-TEE ecosystem. That is inevitable  as some stakeholders are interested, and perhaps more will be in the future. If the best way for this is the OP-TEE SW integration method and thus this change as-is, is questionable.
Going for the full solution (including TS integration) would have the following benefits:
- Simplified adoption for anyone used to OP-TEE development methodology. These stakeholders could use familiar techniques and workflow. I am not familiar to the ecosystem enough to understand the value of this, people might prefer yocto or something else instead.
- Enabling OP-TEE SPMC tests with "real world loads" would be valuable for the OP-TEE project. For that TS integration is a possible way.

> How will it be maintained?

We are using the op-tee SW integration method for daily development of PSA SPs, and thus we have first class interest in keeping this in a good shape. As long as this does not change, it is natural for us to maintain the TS integration and OP-TEE SPMC related content for platforms we use. This currently means SPMC and TS related parts targeting the base FVP.
If that changes and we migrate to e.g. yocto then the value of this becomes questionable. If running TS SPs in the OP-TEE CI will become a thing, then maintaining that and this integration as a dependency my remain our responsibility. If someone will add support for another platform maintaining that will probably be beyond our powers.
We have no plans ready to abandon the OP-TEE integration based workflow, but mainly due to TS CI efficiency issues, we are investigating the topic. But even if we found something better for the TS CI, for daily development we may still stick with these makefiles.

> Can't all this be put in the normal fvp.xml instead?

Perhaps it could, but making the FVP platform TS and SPMC specific might be bad for existing users of the platform. That might mean having some ugly conditional directives in the makefile. Adding a layer on top might be a cleaner approach and a pattern easier to be followed with other platforms in the future.
This does not mean we are not open to changing the structure if you think this is needed.

gyuri-szing:
> Perhaps it could, but making the FVP platform TS and SPMC specific might be bad for existing users of the platform. That might mean having some ugly conditional directives in the makefile. Adding a layer on top might be a cleaner approach and a pattern easier to be followed with other platforms in the future. This does not mean we are not open to changing the structure if you think this is needed.
Apologies, I made a mistake here! This is a manifest change and I am talking about makefiles. Please ignore that. The point I try to make is still valid. SPMC and TS integration is a platform config option, and making a platform wide change is not the right option.

balint-dobszay-arm:
> I think inherit and include this is fine as such, but could you please restructure it a bit so it follow what is stated here? https://optee.readthedocs.io/en/latest/building/gits/manifest.html#sections

I refactored the file, now it should be in sync with these requirements.

jbech-linaro:

balint-dobszay-arm:
> `Reviewed-by: Joakim Bech <joakim.bech@linaro.org>`

Thanks, applied tag and rebased on master.

fvp-ts.xml
```diff
@@ -0,0 +1,27 @@
...
+        <!-- linaro-swg gits -->
+        <!-- Replace Linux with mainline version -->
+        <remove-project path="linux"                    name="linaro-swg/linux.git" />
+        <project        path="linux"                    name="torvalds/linux.git"                       revision="refs/tags/v5.16" clone-depth="1" />
+
+        <!-- Misc gits -->
+        <!-- The fTPM is not used in this config -->
+        <remove-project path="ms-tpm-20-ref"            name="microsoft/ms-tpm-20-ref" />
+        <!-- Add Trusted Services Linux drivers -->
+        <project        path="linux-ffa-tee"            name="linux-arm/linux-trusted-services"         revision="refs/tags/tee-v1" clone-depth="1" remote="arm-gitlab" />
```
jenswi-linaro, line 23: Should it be `linux-arm/linux-trusted-services` or `linux-arm/linux-trusted-services.git`?

jbech-linaro: I think both will work, having said that I think we'v always been explicit in saying `.git`. In addition to that I also think that we've always used the same `path` and `name`, i.e., so you'll see the true name of the git instead of an alternative name.

jenswi-linaro: OK, let me rephrase. Please make it `linux-arm/linux-trusted-services.git` unless there's a good reason to skip the `.git` suffix here.

balint-dobszay-arm: Fixed

balint-dobszay-arm:
> Reviewed-by: Jens Wiklander <jens.wiklander@linaro.org>

Thanks, applied tag and rebased on master.

balint-dobszay-arm:
A review comment on the related build repo patch (https://github.com/OP-TEE/build/pull/571#discussion_r895986648) suggested to rename the Linux FF-A driver targets and paths.

I updated this patch to get in sync with the path renaming. No major change, only `linux-ffa-debugfs` -> `linux-arm-ffa-user` and `linux-ffa-tee` -> `linux-arm-ffa-tee`.

balint-dobszay-arm:
@jenswi-linaro are there any pending issues with this PR? Could we get it merged? Thanks!

jforissier:
@balint-dobszay-arm @jenswi-linaro sorry, I missed that one!

PR #218, opened 2022-09-15
versal: initial manifest

ldts:
Initial manifest to build the Versal ACAP platform.

Tested on the Versal ACAP vck190.

While upstreaming progresses, two of the repositories reference Foundries.io so users can validate all the platform features (eFuse, TRNG, crypto, PUF etc)

It is the expectation that over the coming months - as merges take place - these two repositories will point to the AMD/Xilinx and OP-TEE projects instead.

Signed-off-by: Jorge Ramirez-Ortiz <jorge@foundries.io>

### DIFF
--- a/versal.xml
+++ b/versal.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<manifest>
+	<remote name="github"   fetch="https://github.com" />
+	<default remote="github" revision="master" />
+
+	<!-- OP-TEE gits -->
+	<project path="optee_client"		name="OP-TEE/optee_client.git" />
+	<project path="optee_docs"		name="OP-TEE/optee_docs.git" />
+	<project path="optee_test"		name="OP-TEE/optee_test.git" />
+	<project path="build"			name="OP-TEE/build.git" >
+		<linkfile src="versal.mk" dest="build/Makefile" />
+	</project>
+
+	<!-- linaro-swg gits -->
+	<project path="optee_benchmark"		name="linaro-swg/optee_benchmark.git" />
+	<project path="optee_examples"		name="linaro-swg/optee_examples.git" />
+
+	<!-- Foundries.io gits - upstream work in progress -->
+	<project path="arm-trusted-firmware"	name="foundriesio/arm-trusted-firmware.git"  revision="refs/heads/xlnx-v2.7-versal"/>
+	<project path="optee_os"		name="foundriesio/optee_os.git"              revision="refs/tags/mp-88-3.18+fio"/>
+
+	<!-- Xilinx gits -->
+	<project path="u-boot"			name="Xilinx/u-boot-xlnx.git"	revision="refs/tags/xilinx-v2022.1"/>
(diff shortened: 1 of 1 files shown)


ldts:
maybe I should explicitly point to the 3.18 branch of OP-TEE repos?

jbech-linaro:
> maybe I should explicitly point to the 3.18 branch of OP-TEE repos? while the upstreaming of the FIO repos takes place that is....

I think this should be tracking upstream, we'll tag all manifests when we do the stable releases.

jbech-linaro:

ldts:
> > maybe I should explicitly point to the 3.18 branch of OP-TEE repos? while the upstreaming of the FIO repos takes place that is....
> 
> I think this should be tracking upstream, we'll tag all manifests when we do the stable releases.

sorry what do you mean? do I need to change the Foundries.io repos? this PR uses two Foundries.io repositories: 1) OP-TEE 3.18 + Versal full set and 2) TF-A v2.7 with a single/one-liner patch to jump to OP-TEE (this workaround/hack is required for the time being)...can we then keep it as is for the time being?

jbech-linaro:
> > > maybe I should explicitly point to the 3.18 branch of OP-TEE repos? while the upstreaming of the FIO repos takes place that is....
> > 
> > 
> > I think this should be tracking upstream, we'll tag all manifests when we do the stable releases.
> 
> sorry what do you mean? do I need to change the Foundries.io repos? this PR uses two Foundries.io repositories: 1) OP-TEE 3.18 + Versal full set and 2) TF-A v2.7 with a single/one-liner patch to jump to OP-TEE (this workaround/hack is required for the time being)...can we then keep it as is for the time being?

This is good as it is, that is what I was trying to say (hence my R-B tag). Hopefully it'll be possible to switch over to upstream repo instead of Foundries.io repos on the future. My comment about tagging was more of a generic comment. I.e., the other ones that are prefixed with "optee_" etc. Those will get a stable tag once we make the normal OP-TEE release.

ldts:
updated with tags

ldts:
> > > > maybe I should explicitly point to the 3.18 branch of OP-TEE repos? while the upstreaming of the FIO repos takes place that is....
> > > 
> > > 
> > > I think this should be tracking upstream, we'll tag all manifests when we do the stable releases.
> > 
> > 
> > sorry what do you mean? do I need to change the Foundries.io repos? this PR uses two Foundries.io repositories: 1) OP-TEE 3.18 + Versal full set and 2) TF-A v2.7 with a single/one-liner patch to jump to OP-TEE (this workaround/hack is required for the time being)...can we then keep it as is for the time being?
> 
> This is good as it is, that is what I was trying to say (hence my R-B tag). Hopefully it'll be possible to switch over to upstream repo instead of Foundries.io repos on the future. My comment about tagging was more of a generic comment. I.e., the other ones that are prefixed with "optee_" etc. Those will get a stable tag once we make the normal OP-TEE release.

yes. I understood your point after I asked the question. ok cool. thanks

ldts:
will send the doc update in a bit.

PR #230, opened 2023-01-13
Fix make_stable.sh

### LINKED ISSUE #226: Syntax error in stable release
Closing backslash is missing from XML entry resulting in repo tool failure in 3.19 (stable) release.
See: https://github.com/OP-TEE/manifest/blob/3.19.0/fvp-ts.xml#L9

Correct line would be: **<remove-project path="build"  name="OP-TEE/build.git" revision="refs/tags/3.19.0" clone-depth="1"/>**

"revision" and "clone-depth" attributes should also be removed from remove-project element. It is not erronous, but does not makes sense IMO.

balint-dobszay-arm:
Currently the script doesn't have filtering based on which type of element the line contains. This can lead to unexpected results or syntax errors, e.g. when using a "remove-project" element. Fix this issue by introducing an address to the sed expressions which narrows the scope down to only "project" type lines.
Resolves #226

### DIFF
--- a/make_stable.sh
+++ b/make_stable.sh
@@ -44,15 +44,15 @@ do
 		FILE=$xml
 	fi
 
-	cat $xml | 
-		sed "s/\(OP-TEE\/.*\)revision.*/\1\/>/" | # Removes old revision
-		sed "s/\(OP-TEE.*\"\)/\1 revision=\"refs\/tags\/${VERSION}\" clone-depth=\"1\"/" |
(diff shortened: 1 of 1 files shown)


jforissier (approved):
LGTM, thanks for the patch!

balint-dobszay-arm:
Sorry I forgot about this, can we get this merged please?

jforissier:
> Sorry I forgot about this, can we get this merged please?

Me too! :open_mouth: Merging now. Thanks.

PR #231, opened 2023-01-16
fvp: Update TF-A version to v2.8 for fvp-ts target

imre-kis-arm:
Also update Mbed-TLS to v2.28.1 to satisfy the requirements of the new TF-A version.

### DIFF
--- a/fvp.xml
+++ b/fvp.xml
@@ -23,8 +23,8 @@
         <project path="edk2"                 name="tianocore/edk2.git"                    revision="dd4cae4d82c7477273f3da455084844db5cca0c0" />
         <project path="edk2-platforms"       name="tianocore/edk2-platforms.git"          revision="02daa58c21f89628b4d8c76f95f3a554289149bc" />
         <project path="grub"                 name="grub.git"                              revision="refs/tags/grub-2.02" clone-depth="1" remote="savannah" />
-        <project path="trusted-firmware-a"   name="TF-A/trusted-firmware-a.git"           revision="refs/tags/v2.6" clone-depth="1" remote="tfo" />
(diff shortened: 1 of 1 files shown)


fvp-ts.xml
```diff
@@ -17,6 +17,14 @@
...
 
         <!-- Misc gits -->
+        <!-- Update to TF-A v2.8 -->
+        <remove-project path="trusted-firmware-a"       name="TF-A/trusted-firmware-a.git" />
+        <project path="trusted-firmware-a"              name="TF-A/trusted-firmware-a.git"              revision="refs/tags/v2.8" clone-depth="1" remote="tfo" />
+
+        <!-- TF-A v2.8 requires Mbed-TLS v2.28.1 -->
+        <remove-project path="mbedtls"                  name="Mbed-TLS/mbedtls.git" />
+        <project path="mbedtls"                         name="Mbed-TLS/mbedtls.git"                     revision="refs/tags/mbedtls-2.28.1" />
+
```
jforissier, line 27: Why not update `fvp.xml` instead?

imre-kis-arm: When I've first tested the TF-A & MbedTLS update by changing fvp.xml the regular, non-SPMC mode (build/fvp.mk) build failed with the `MEASURED_BOOT=y` configuration. I thought this is because of the update but I didn't have the capacity to investigate it in detail. This is why I decided to limit the scope and only change fvp-ts.xml.
Now I've checked it and it turned out that it already fails with the original TF-A & MbedTLS version, so this is a separate issue.
Using the updated config the build works and xtests passes when `MEASURED_BOOT=n`.
If it's OK for you, I move this change to fvp.xml as suggested and open an issue in the build repo about the measured boot build failure.

jforissier: @imre-kis-arm sounds good to me

jforissier (commented):

PR #232, opened 2023-01-26
AMD/Xilinx Versal ACAP: update to upstream tree

ldts:
All xtests pass with 3.20

Signed-off-by: Jorge Ramirez-Ortiz <jorge@foundries.io>

### DIFF
--- a/versal.xml
+++ b/versal.xml
@@ -4,6 +4,7 @@
 	<default remote="github" revision="master" />
 
 	<!-- OP-TEE gits -->
+	<project path="optee_os"		name="OP-TEE/optee_os.git" />
 	<project path="optee_client"		name="OP-TEE/optee_client.git" />
 	<project path="optee_docs"		name="OP-TEE/optee_docs.git" />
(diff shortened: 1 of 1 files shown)


ldts:
@nathan-menhorn just a heads up, after the 3.20 update we can/should switch to upstream.

ldts:
@jforissier please could you pick this?

jenswi-linaro:

ldts:
thanks, having forks is a sign of failure so glad we got rid of it...

nathan-menhorn:
> @nathan-menhorn just a heads up, after the 3.20 update we can/should switch to upstream.
> 
> I validated with https://optee.readthedocs.io/en/latest/building/devices/versal.html
> 
> One thing to notice is that in upstream, defining CFG_VERSAL_FPGA_DDR_ADDR is enough for the TEE to attempt to program the FPGA (before we used to have an additional knob..FPGA_INIT or something similar). Just in case your builds are pointing to some FPGA firmware that might lock the kernel.

Hi @ldts thanks for the heads up. Please send me a a ping on email or slack when the factory has switched over.

PR #236, opened 2023-04-13
fvp-ts: pin TS component versions

imre-kis-arm:
Set fixed version for linux-arm-ffa-user, linux-arm-ffa-tee and trusted-services components in fvp-ts.xml.

### DIFF
--- a/fvp-ts.xml
+++ b/fvp-ts.xml
@@ -3,6 +3,7 @@
         <include name="fvp.xml" />
 
         <remote name="arm-gitlab" fetch="https://git.gitlab.arm.com" />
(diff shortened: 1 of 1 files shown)


jenswi-linaro (commented):

imre-kis-arm:
This PR was opened before the previous release in order to pin the TS component versions. This way the released manifest would have fixed and compatible versions of both OP-TEE and TS components.
Although the PR didn't end up in the previous release it is still relevant and we would like to have it in the upcoming release. I've updated the TS components to latest versions.

jforissier:
That's fine, I am merging this. Thanks.

PR #239, opened 2023-05-08
stm32mp1: upgrade u-boot version, fetch edk2/edk2-platform and mbedtls

etienne-lms:


jforissier (commented):
In the description for "stm32mp1: fetch mbedtls for tf-a authenticated boot":
s/crypro/crypto/

Other than that:

(preferably without the Change-Id: tags)

PR #240, opened 2023-07-03
imx: use github instead of codeaurora

jenswi-linaro:
source.codeaurora.org is closed, use https://github.com/nxp-imx instead.


jforissier (commented):

jenswi-linaro:
Tag applied.

PR #254, opened 2023-11-17
qemu_v8: add Xen repository using tag RELEASE-4.18.0

jenswi-linaro:
Adds Xen repository to be used when compiling Xen, both the hypervisor itself and the user-space parts built with Buildroot.


jforissier (commented):

jenswi-linaro:
Tag applied, removed the `Change-Id` also.

PR #261, opened 2024-01-10
qemu_v8: rust: Uprev to no-std branch tip

b49020:
OP-TEE rust SDK build framework have been refactored to support 32-bit as well as 64-bit TAs build. Along with that mixed build options support is added where host applications build for 64-bit works with TAs build for 32-bit or vice-versa.


b49020:
Tag applied, thanks.

etienne-lms:
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>`

b49020:
Tag applied, thanks.

PR #262, opened 2024-01-15
Refactor manifests/avoid duplication

igoropaniuk:
1. Use common repositories defined in common.xml manifest to avoid
    code duplication.
2. Drop manifest for Verdin iMX8M Mini, as it's not maintained anymore.


jenswi-linaro:
I'd rather wait with an eventual refactoring of the manifest until after the release.

jforissier:
> I'd rather wait with an eventual refactoring of the manifest until after the release.

Agreed. As usual, and since -rc1 has been tagged already, I will only merge fixes.

jbech-linaro:
I'm OK with these changes and it's a nice clean up. But at the same time, if we for any reason in the future would need to move something to a specific git commit for a particular device, then we cannot do that as easy any longer. For that you'd have to use a local manifest, that works locally. However, historically we haven't had that happen a lot, so again, I'm OK with this, so please use my RB tag.

jenswi-linaro:
You're supposed to be able to override what's in a previously included file. See for example `fvp-ts.xml`.

jbech-linaro:
> You're supposed to be able to override what's in a previously included file. See for example `fvp-ts.xml`.

OK, then I have no concerns. Thanks!

am43xx.xml
```diff
@@ -5,19 +5,12 @@
 
         <default remote="github" revision="master" />
```
jenswi-linaro, line 6: Could this be taken from `common.xml` instead?

etienne-lms (commented):
`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commits
"verdin: drop manifest" and
"include: move common repositories to separate file".

For commit "common: use common manifest to avoid code duplication":
commit message could mention that reference to optee_benchmark repository is removed as the feature is also currently being removed. Or you could have a dedicated commit for removing these refs.

etienne-lms:
My bad, optee_benchmark is listed in _common.xml_.

`Acked-by: Etienne Carriere <etienne.carriere@foss.st.com>` for commit
"common: use common manifest to avoid code duplication".

igoropaniuk:
@etienne-lms I'll drop optee_manifest in a separate PR

am43xx.xml
igoropaniuk: fixed

igoropaniuk:
Addressed @jenswi-linaro comment, applied tags from @jbech-linaro @etienne-lms , rebased

common.xml
```diff
@@ -0,0 +1,18 @@
...
+        <project path="optee_os"             name="OP-TEE/optee_os.git" />
+        <project path="optee_test"           name="OP-TEE/optee_test.git" />
+
+        <!-- linaro-swg gits -->
+        <project path="linux"                name="linaro-swg/linux.git"                  revision="optee" clone-depth="1" />
+        <project path="optee_benchmark"      name="linaro-swg/optee_benchmark.git"/>
+        <project path="optee_examples"       name="linaro-swg/optee_examples.git" />
+
+        <!-- buildroot git -->
+        <project path="buildroot"            name="buildroot/buildroot.git"               revision="refs/tags/2022.11.1" clone-depth="1" remote="github" />
```
jenswi-linaro, line 17: The default remote is "github" already.

versal.xml
```diff
@@ -1,20 +1,11 @@
...
-	<!-- OP-TEE gits -->
-	<project path="optee_os"		name="OP-TEE/optee_os.git" />
-	<project path="optee_client"		name="OP-TEE/optee_client.git" />
-	<project path="optee_docs"		name="OP-TEE/optee_docs.git" />
-	<project path="optee_test"		name="OP-TEE/optee_test.git" />
-	<project path="build"			name="OP-TEE/build.git" >
-		<linkfile src="versal.mk" dest="build/Makefile" />
+	<include name="common.xml" />
+	<project path="build"                   name="OP-TEE/build.git" >
+			<linkfile src="qemu_v8.mk" dest="build/Makefile" />
```
jenswi-linaro, line 5: s/qemu_v8.mk/versal.mk/ ?

common.xml
igoropaniuk: fixed, thanks!

versal.xml
igoropaniuk: fixed

jenswi-linaro:

igoropaniuk:
@jenswi-linaro tag applied. I've just stumbled upon another issue, rpi3 was using `optee` branch instead of `rpi3-optee-6.7`, this is fixed already

PR #263, opened 2024-01-15
poplar: switch to latest u-boot master

igoropaniuk:
Switch to the latest U-Boot master which includes these fixes:
- 66a3618b9a poplar: provide more space for kernel image
- 01201dbd3b poplar: use random mac address

### DIFF
--- a/poplar.xml
+++ b/poplar.xml
@@ -21,7 +21,7 @@
         <!-- 96boards-poplar gits -->
         <project path="l-loader"             name="96boards-poplar/l-loader.git"          revision="master" clone-depth="1" />
         <project path="poplar-tools"         name="96boards-poplar/poplar-tools.git"      revision="master" clone-depth="1" />
-        <project path="u-boot"               name="u-boot/u-boot.git"                     revision="refs/tags/v2021.07" clone-depth="1" remote="denx" />
+        <project path="u-boot"               name="u-boot/u-boot.git"                     revision="697758e7c81131da6db0e3b10515019fe3aca8c9" clone-depth="1" remote="denx" />
 
         <!-- Misc gits -->
         <project path="buildroot"            name="buildroot/buildroot.git"               revision="refs/tags/2022.11.1" clone-depth="1" />


poplar.xml
```diff
@@ -21,7 +21,7 @@
         <!-- 96boards-poplar gits -->
         <project path="l-loader"             name="96boards-poplar/l-loader.git"          revision="master" clone-depth="1" />
         <project path="poplar-tools"         name="96boards-poplar/poplar-tools.git"      revision="master" clone-depth="1" />
-        <project path="u-boot"               name="u-boot/u-boot.git"                     revision="refs/tags/v2021.07" clone-depth="1" remote="denx" />
+        <project path="u-boot"               name="u-boot/u-boot.git"                     revision="697758e7c81131da6db0e3b10515019fe3aca8c9" clone-depth="1" remote="denx" />
```
jenswi-linaro, line 24: Isn't there any tag that can work for poplar?

jbech-linaro: +1 for Jens suggestion. Also, this as written here will not work, since `clone-depth="1"` doesn't work with pure commits, please see [here](https://optee.readthedocs.io/en/latest/building/gits/manifest.html#when-to-use-clone-depth-1) for details about that.

jforissier: > Also, this as written here will not work, since `clone-depth="1"` doesn't work with pure commits, please see [here](https://optee.readthedocs.io/en/latest/building/gits/manifest.html#when-to-use-clone-depth-1) for details about that.

This issue was apparently fixed by commit https://android.googlesource.com/tools/repo/+/6c5944606af3c52626745b5256e9e5f0d5c362a2.

igoropaniuk: @jenswi-linaro Not really, the fixes were not included in u-boot `v2024.01` release, expected to be in the next `v2024.04`

jbech-linaro: Ah, cool, then we should update our docs. Thanks!

igoropaniuk: @jenswi-linaro so let me know, should we wait till v2024.04 release or we can proceed with hash for now?
Anyway boot is broken now already

jenswi-linaro: Let's take this now.

jenswi-linaro:

igoropaniuk:
@jenswi-linaro tag applied

PR #268, opened 2024-01-24
zynqmp.xml: add clone-depth="1"

jforissier:
Add clone-depth="1" to all the ZynqMP projects to make the initial repo sync faster. In my experience, with this the sync time goes down from 10 minutes to about 1 minute.


jenswi-linaro (commented):

PR #270, opened 2024-02-06
Update Buildroot to 2023.11.1

jforissier:
Update Buildroot to the latest released version. One nice side effect is that the source repository for the ncurses package has changed to a more reliable location (I have seen quite a few network timeouts and retries in CI recently, causing a job to exceed 4 hours of build time).


jforissier:
Depends on https://github.com/OP-TEE/build/pull/732.

jenswi-linaro:

PR #273, opened 2024-03-18
Manifest cleanup

fboudra:
Some manifests clean up in an attempt to standardize the style and make it easier to compare/use.

### DIFF
--- a/common.xml
+++ b/common.xml
@@ -2,7 +2,7 @@
 <manifest>
   <remote name="github" fetch="https://github.com"/>
 
-  <default remote="github" revision="master"/>
+  <default remote="github" revision="master" sync-j="4" clone-depth="1"/>
 
   <!-- OP-TEE gits -->
   <project path="optee_client" name="OP-TEE/optee_client.git"/>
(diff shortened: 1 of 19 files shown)


jbech-linaro:
Always nice with clean-ups, but please have a look at [this page](https://optee.readthedocs.io/en/latest/building/gits/manifest.html#project-xml-elements) as well to make sure that the information there still is up-to-date as well. We did try to use that as a guidance in the past to make them look the same, to make it easier to diff etc. . I.e., I see that after merging this, we need to update the docs page accordingly.

fboudra:
@jbech-linaro thanks for the pointer. I'll have a look if we can automate this to check the guidance is followed or propose a doc update if it make sense.

common.xml
```diff
@@ -2,17 +2,17 @@
 <manifest>
   <remote name="github" fetch="https://github.com"/>
 
-  <default remote="github" revision="master"/>
+  <default remote="github" revision="master" sync-j="4" clone-depth="1"/>
```
jenswi-linaro, line 5: What effect does this have on, for instance, the optee_os git? Will it also have a clone-depth of 1?
We'd like to clone the entire git history for all the optee* gits.

jenswi-linaro: When a different `-j` flag is passed to repo, will that take precedence?

PR #275, opened 2024-03-21
qemu_v8: SCP-firmware repository relocation

vingu-linaro:
SCP-firmware has moved from github to Arm gitlab

### DIFF
--- a/qemu_v8.xml
+++ b/qemu_v8.xml
@@ -3,6 +3,7 @@
         <remote name="tfo"      fetch="https://git.trustedfirmware.org" />
         <remote name="u-boot"   fetch="https://source.denx.de/u-boot" />
         <remote name="xenbits"  fetch="https://xenbits.xen.org/" />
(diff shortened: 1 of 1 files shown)


qemu_v8.xml
```diff
@@ -3,6 +3,7 @@
         <remote name="tfo"      fetch="https://git.trustedfirmware.org" />
         <remote name="u-boot"   fetch="https://source.denx.de/u-boot" />
         <remote name="xenbits"  fetch="https://xenbits.xen.org/" />
+        <remote name="gitlab"   fetch="https://git.gitlab.arm.com/" />
```
jforissier, line 6: Since this is not "the" GitLab (gitlab.com) I suggest `name="arm-gitlab"`

jforissier (commented):
One comment then:

qemu_v8.xml
vingu-linaro: Yeah, I was hesitant between the 2 names but chose the shortest one. I'm going to use arm-gitlab and add your ack

PR #279, opened 2024-05-20
fvp: Add hafnium component to enable running wih SPMC at EL2

sudeep-holla:
Let us add the hafnium component in order to enable building and testing SPMC at EL2. Also update TF-A to match Hafnium version at v2.10

### DIFF
--- a/fvp.xml
+++ b/fvp.xml
@@ -8,7 +8,7 @@
                 <linkfile src="fvp.mk" dest="build/Makefile" />
         </project>
 
-        <project path="trusted-firmware-a"   name="TF-A/trusted-firmware-a.git"           revision="refs/tags/v2.9" clone-depth="1" remote="tfo" />
+        <project path="trusted-firmware-a"   name="TF-A/trusted-firmware-a.git"           revision="refs/tags/v2.10" clone-depth="1" remote="tfo" />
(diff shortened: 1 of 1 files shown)


jforissier:
I'd rather have two separate commits. In any case:

sudeep-holla:
I can do that if that is the preference, I was being lazy I guess :).

jforissier:
Well no big deal but having both changes in one patch makes it unclear if there is a dependency here, i.e., does the hafnium config require TF-A v2.10 or is v2.9 still OK? If not you should first add Hafnium then update TF-A. If yes, then do the opposite and mention it in the commit description ("fvp: update TF-A to v2.10": "In order to enable running with SPMC at EL2 with Hafnium in a subsequent commit, update TF-A to v2.10").

sudeep-holla:
> Well no big deal but having both changes in one patch makes it unclear if there is a dependency here, i.e., does the hafnium config require TF-A v2.10 or is v2.9 still OK? If not you should first add Hafnium then update TF-A. If yes, then do the opposite and mention it in the commit description ("fvp: update TF-A to v2.10": "In order to enable running with SPMC at EL2 with Hafnium in a subsequent commit, update TF-A to v2.10").

I did the latter. In general, they are features  set compatible. Not sure if we can run hafnium at v2.10 keeping TF-A at v2.9, but some FF-A features got added in both and are needed to get the feature enabled in end-to-end stack. That is the reason why I bumped it to v2.10

PR #284, opened 2024-06-27
qemu, poplar: get U-Boot from source.denx.de rather than gitlab.denx.de

jforissier:
THe official domain name for U-Boot source code is source.denx.de not gitlab.denx.de. Update QEMU and Poplar accordingly. Other platforms are already using the correct URL.

Link: https://docs.u-boot.org/en/v2024.04/build/source.html [1]


jenswi-linaro:

PR #291, opened 2024-08-16
qemu_v8: replace trustedfirmware.org repositories with their GitHub mirrors

jforissier:
git.trustedfirmware.org has been unreliable and has caused CI errors a couple of times so switch to the GitHub mirrors instead. Another advantage is that GitHub is a bit faster.

Link: https://github.com/OP-TEE/optee_os/pull/6880#issuecomment-2160979548
Link: https://github.com/OP-TEE/optee_os/pull/6993#issuecomment-2290665315


jenswi-linaro (commented):

PR #293, opened 2024-09-25
Move ms-tpm-20-ref into common.mk

jenswi-linaro:
Move ms-tpm-20-ref from fvp.xml into common.mk to make the reference source available in all configurations.

Change-Id: I6816c52ed47dcff3b2df92d209161033ccd8d453


jbech-linaro:
I see it's a draft, but still

jenswi-linaro:
Replaced by https://github.com/OP-TEE/manifest/pull/298

PR #296, opened 2024-10-11
fvp-ts: Update Trusted Services

imre-kis-arm:
Update Trusted Services to the latest version on integration branch as of 2024-10-11.
The patch also removes the trusted-firmware-a and mbedtls entries from fvp-ts.xml because the included fvp.mk now has the same versions of these components.


imre-kis-arm:
Added `Acked-by` tag.

PR #305, opened 2024-12-16
qemu_v8: upgrade QEMU to v9.2.0

jenswi-linaro:
Upgrade QEMU to v9.2.0. This is needed to avoid error with the latest kernel like:
```
ERROR:../target/arm/internals.h:760:regime_is_user: code should not be reached Bail out!
ERROR:../target/arm/internals.h:760:regime_is_user: code should not be reached make: 
*** [Makefile:571: run-only] Aborted (core dumped)
```

Depends on https://github.com/OP-TEE/build/pull/798


jforissier (commented):

jenswi-linaro:
Tag applied.

PR #312, opened 2025-02-07
versal_net: fix typo in bootgen tag

jcorbier:
Bootgen Git tag was incorrectly copied from the Linux one (xilinx-v2024.2) whereas it should be xilinx_v2024.2.

Sorry about that...


jforissier (commented):

PR #316, opened 2025-04-01
qemu_v8: upgrade Xen to RELEASE-4.20.0

jenswi-linaro:
Upgrade Xen to RELEASE-4.20.0. This fixes FF-A version negotiation so it can downgrade to version 1.1 properly if version 1.2 is encountered in firmware, for instance, OP-TEE.


jforissier (commented):

PR #317, opened 2025-04-03
qemu_v8: upgrade scp firmware to v2.16.0

vingu-linaro:
Upgrade SCP-firmware version to v2.16.0 which has been released mid march and includes the fix for the scp notification compilation among other things.


jenswi-linaro:

PR #322, opened 2025-04-30
qemu_v8: hafnium: synchronize submodules on checkout

jforissier:
When doing the "repo sync", Hafnium needs to be checked out in full including its submodules (prebuilt etc.) because it won't do it automatically when building. To do that, add sync-s="true".

Fixes: 9dd49cf8e39b ("qemu_v8: upgrade hafnium to v2.12")


jenswi-linaro:
I thought the `.hafnium_checkout` target in the build git took care of that.

jforissier:
True. The problem is that I manually checked out Hafnium v2.12 (from an older checkout) without doing "git submodule update", and that's a mistake. Closing this.

PR #324, opened 2025-05-05
qemu_v8: clone Xen from its GitHub mirror

jforissier:
Use https://github.com/xen-project/xen.git rather than https://xenbits.xen.org/git-http/xen.git for reliability reasons. From the OP-TEE OS CI logs:

 repo has been initialized in /__w/optee_test/optee_repo_qemu_v8
 error: Cannot fetch git-http/xen.git from https://xenbits.xen.org/git-http/xen.git
 fatal: unable to access 'https://xenbits.xen.org/git-http/xen.git/': Failed to connect to xenbits.xen.org port 443 after 135097 ms: Could not connect to server

### DIFF
--- a/qemu_v8.xml
+++ b/qemu_v8.xml
@@ -2,7 +2,6 @@
 <manifest>
         <remote name="tfo"    fetch="https://git.trustedfirmware.org" />
         <remote name="u-boot"   fetch="https://source.denx.de/u-boot" />
-        <remote name="xenbits"  fetch="https://xenbits.xen.org/" />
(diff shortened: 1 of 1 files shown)


jforissier:
> Seems like GitHub is the trustworthy and stable one these days! `Reviewed-by: Joakim Bech <joakim.bech@linaro.org>`

Yes. I have very rarely seen it fail, so IMO it should be our preferred source for third party projects for CI (qemu_v8/qemu).

PR #339, opened 2025-12-08
qemu_v8: use sync-s="true" for trusted-firmware-a

jforissier:
Trusted Firmare-A v2.14 needs to be cloned with its submodules properly initialized. For this, add sync-s="true" (which is equivalent to running 'git submodule update --init' after cloning). Fixes CI errors in the "FW handoff" and "SMPC_AT_EL=2" QEMUv8 jobs.


jenswi-linaro:

PR #341, opened 2026-02-04
qemu_v8: rust: pin the latest version

ivila:
Resolve CI failures in OP-TEE/optee_os#7690 by ensuring the linker script is correctly preprocessed during the build phase.

### DIFF
--- a/qemu_v8.xml
+++ b/qemu_v8.xml
@@ -8,7 +8,7 @@
         </project>
 
         <!-- Misc gits -->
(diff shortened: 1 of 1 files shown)


jforissier:

But I wouldn't mind having two commits if you want to set `clone-depth="1"` (which is a good idea I think).

jenswi-linaro:

ivila:
I have split the changes into two distinct commits and added the Acked-by trailers.

PR #345, opened 2026-04-09
imx: switch IMX TF-A fork and version v2.12

jenswi-linaro:
Switch IMX TF-A fork and update version to lf-6.18.2-1.0.0 based on TF-A version 2.12.

Depends on https://github.com/OP-TEE/build/pull/858

### DIFF
--- a/imx.xml
+++ b/imx.xml
@@ -9,7 +9,7 @@
         </project>
 
         <!-- Misc gits -->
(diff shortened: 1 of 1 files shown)


jenswi-linaro:
Please review so we can include it in the release.

jforissier:
FWIW:

jenswi-linaro:
Tag applied

PR #349, opened 2026-06-12
nanopc-t6: add rkbin project and TPL linkfile

edtubbs:
Add the rkbin project and link the NanoPC-T6 TPL binary into out/ for the build.


jforissier:

